PR #3, opened 2024-08-01
Add autogen for Genkernel, without touching existing static ebuild.

cuantar:
Genkernel has features still unsupported by `ramdisk`, and the latter can probably be expected to be unmaintained unless we adopt it.  I'd like to add Genkernel to our tree, for now, until we decide what to do.


cuantar:
new version didn't merge cleanly due to a changed/missing file; will fix and update the PR

cuantar:
This is the version that I've been using for a while.

PR #8, opened 2024-08-01
remove old static ebuild for genkernel

cuantar:


cuantar:
In Funtoo, this ebuild was hardmasked for removal because it no longer works with the lvm present in the tree.

PR #26, opened 2024-08-02
[games-kit] Odamex: improve template

cuantar:
This change probably won't fix the autogen issue (seems to be a metatools bug) but will clean up the generated ebuilds.


org-tekeli-borisp:
Regen started. We will see

org-tekeli-borisp:
tree regeneration successful

PR #28, opened 2024-08-02
Correct verson handling for patchlevel

cuantar:
The patches only need to be applied to versions like `8.2`, and now (but perhaps not previously) there is a tarball for `8.2.13`, the current version after patching.  The `autogen.py` seemingly did not expect this case.


cuantar:
Does anybody know how to link this PR to the issue in mark-issues?

cuantar:
Fixes macaroni-os/mark-issues#16.

PR #29, opened 2024-08-02
[core-kit] add app-text/asciidoc to build-time deps for genkernel

cuantar:
Genkernel does not install on machines that don't have `a2x` (from `asciidoc`) present.


cuantar:
@geaaru @org-tekeli-borisp should be safe to merge this, it is trivial.

PR #43, opened 2024-08-04
[core-kit] [portage] change GENTOO_MIRRORS -> distfiles.macaronios.org

cuantar:
Also, remove old version and bump revision of current version.

This should be safe to merge.   I will leave it for someone else to do the
honors.


cuantar:
Closes macaroni-os/mark-issues#20.

PR #49, opened 2024-08-05
 [bugfix] [kernel] Repair the debian-sources and genkernel ebuilds

cuantar:
Previous PR got clobbered by my tools that were unintentionally aggressive about repository state ownership, when the stable `kit-fixups` apparently was force-pushed on top of my repository.  So, here it is again!  Hopefully here to stay.


cuantar:
Successfully installed on a variety of machines and in new metro stages.

core-kit/next/sys-kernel/debian-sources/debian-sources-6.9.12_p1.ebuild
```diff
@@ -317,15 +317,17 @@ src_install() {
...
+			--ramdisk-modules \
 			--initramfs-filename=initramfs-${KERN_SUFFIX} || \
 				die "genkernel failed:  $?" \
 	)
 
 	# copy the fresh Genkernel cache into the image
 	if use genkernel; then
 		dodir /var/cache/genkernel
-		cp "${WORKDIR}/genkernel-cache/*" "${D}/var/cache/genkernel/" || die
+		cp -r "${WORKDIR}/genkernel-cache/4.3.10" "${D}/var/cache/genkernel/" || die
```
org-tekeli-borisp, line 330: genkernel is autogened now. If we get a new version of genkernel e.g. 4,3.11, could this break again?

core-kit/next/sys-kernel/debian-sources/debian-sources-6.9.12_p1.ebuild
```diff
@@ -317,15 +317,17 @@ src_install() {
 			--kerneldir=${S} \
 			--bootdir=${D}/boot \
 			--cachedir=${WORKDIR}/genkernel-cache \
+			--no-clear-cachedir \
 			--kernel-modules-prefix=${D} \
+			--ramdisk-modules \
```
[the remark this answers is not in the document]
cuantar: Yes, correct.  `$ genkernel --help | grep modules`:
```        --ramdisk-modules       Copy required modules to the initramfs```

core-kit/next/sys-kernel/debian-sources/debian-sources-6.9.12_p1.ebuild
cuantar: It is pinned in the autogen YAML.  It definitely will break again if someone bumps the version without testing it --- this version was the last one without further complications. We are unlikely ever to get another version --- after this I'll do `dracut` and likely make that default because it is well-supported and reliable, and Macaroni is using it too.

PR #36, opened 2024-08-03
My stab on thin-provisioning-tools

cepelinas9000:
My stab on problem and **Please test and review ebuild before merge (my approach can be wrong)**.

 
There are few notes:
1. in rust.py written "Upon completion, metatools will create its own archive from scratch and store it in the local binary object store, and copy it to $DISTDIR for you for convenience. " -that is not very convenient  if you are not cdn
2. didn't notice crate version overrides in autogen [rust.py] (created hard block for io-uring use flag and hope that next tools version have sorted out)
3. bindgen didn't found clang headers from (/usr/lib/clang/16/include) - it can be sys-devel/clang problem
4. project have makefile and manual install method - there is missing function cargo_target_dir (like in gentoo cargo.eclass), which determines build type

as usual

```console
doit --pkg thin-provisioning-tools
INFO     Autogen: sys-block/thin-provisioning-tools (latest)                                                                                                                            
WARNING  dict key compression overwritten.                                                                                                                                              
WARNING  dict key compression overwritten.                                                                                                                                              
WARNING  dict key version overwritten.                                                                                                                                                  
WARNING  dict key version overwritten.                                                                                                                                                  
WARNING  dict key version overwritten.                                                                                                                                                  
WARNING  dict key version overwritten.                                                                                                                                                  
WARNING  dict key version overwritten.                                                                                                                                                  
WARNING  dict key version overwritten.                                                                                                                                                  
INFO     Created: sys-block/thin-provisioning-tools/thin-provisioning-tools-1.0.14.ebuild                                                                                               

as@localhost ~/proj/funtoo/kit-fixups/core-kit/curated $ git status
On branch macaroni_os_mark-issues_2
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	sys-block/thin-provisioning-tools/Manifest
	sys-block/thin-provisioning-tools/thin-provisioning-tools-1.0.14.ebuild
	../../python-modules-kit/next/dev-python/setuptools_scm/

nothing added to commit but untracked files present (use "git add" to track)
as@localhost ~/proj/funtoo/kit-fixups/core-kit/curated $ git push
Everything up-to-date

```

PR #60, opened 2024-08-09
Add thin-provisioning-tools, metatools autogen, and improved debian-sources ebuilds

cuantar:
This PR fixes several bugs and adds needed packages.  It has been tested in a series of two complete forks of M.A.R.K.; this tree from which the PR comes should only see changes that pass a series of tests and don't break the tree or `metro` builds, and generally seem to work for me.

Bugs fixed:
- improper /boot detection in `mount-boot.eclass`
- genkernel cache now works reasonably well
- kernel ebuilds are robust and work everywhere I've tested them now
- `acpi-ec` USE flag in `debian-sources` ebuilds now turns on `CONFIG_DEBUG_FS` that is needed for my fan control program on my laptop to operate correctly, in addition to its previous behavior

New features:
- `metatools` autogen changed from PyPI to the Macaroni OS fork at Github (currently pinned to a commit hash, same one as the ebuild in `geaaru-kit`
- kernel ebuilds now more clearly track the Debian `testing` (`trixie`) and `stable` (`bookworm`) branches
- Funtoo's `debian-sources-6.5.10_p1` is now deprecated; it's in the tree but has `KEYWORDS=""`


cuantar:
There is one more big patch that needs to be part of this PR; still running the last test on it. Stay tuned...

geaaru:
@cuantar i think that you can set metatools hash to last commit too with the @org-tekeli-borisp patch. we using it on regen but i haven't yet aligned this on my overlay.

cuantar:
@geaaru okay, I'll do that!

cuantar:
This PR also fixes #36 and mark-issues/#32 macaroni-os/mark-issues/#46

PR #61, opened 2024-08-10
sys-fs/zfs-kmod changes in the $DEPMOD invocation at the end of kerne…

org-tekeli-borisp:
…l scrips/depmod.sh broke old strategy whereby $DEPMOD was set to ":". See similar https://bugs.gentoo.org/916587

Before
image
After
image


cuantar (approved):
Do you happen to know why this is needed?

org-tekeli-borisp:
@cuantar 
> With kernel commit 4d15c9fa058e6, the "$DEPMOD" invocation at the end of kernel scripts/depmod.sh became an "exec", which broke the old strategy in linux-mod-r1.eclass whereby $DEPMOD was set to ":".  (See also kernel commit 79b96c332241c.)

See https://bugs.gentoo.org/916587

PR #64, opened 2024-08-12
=sys-kernel /linux-headers-6.9 for >=net-misc/networkmanager-1.48.8

org-tekeli-borisp:
Should solve https://github.com/macaroni-os/mark-issues/issues/68


org-tekeli-borisp:
@cuantar  has a better idea: 
> new ebuild that simply unpacks the kernel sources, executes make headers_install, and then finishes. this part of our infrastructure would then be independent of Gentoo and aligned with the kernel we're using

PR #78, opened 2024-08-19
Removing rust extension merged by harvester2024 07 and restore cargo extension

geaaru:


geaaru:
Related to macaroni-os/mark-issues#88

PR #83, opened 2024-08-23
[net-misc] net-misc/dropbox - drop x86 support

geaaru:
Closes: macaroni-os/mark-issues#95


geaaru:
I will cherry-pick on `master` too

PR #85, opened 2024-08-24
[kernel] Fix metro builds by using a more permissive copy command

cuantar:
In `pkg_postinst` , the `genkernel` cache is copied from the image back to the host rootfs.  Using `rsync` for the copy seems to be more resilient than using `cp -r`.  This PR fixes metro builds. 

image

### DIFF
--- a/core-kit/next/sys-kernel/debian-sources/debian-sources-6.1.99_p1-r1.ebuild
+++ b/core-kit/next/sys-kernel/debian-sources/debian-sources-6.1.99_p1-r1.ebuild
@@ -372,12 +372,12 @@ pkg_postinst() {
 	# copy the fresh Genkernel cache into the image, but
 	# only if the host doesn't have a cache already existing.
 	# addread /var/cache/genkernel/4.3.10
-	if use genkernel && [[ -d /var/cache/genkernel/4.3.10 ]]; then
-		einfo "Leaving pre-existing genkernel cache at /var/cache/genkernel/4.3.10 alone."
+	if use genkernel && [[ -d "${ROOT}"cache/genkernel/4.3.10 ]]; then
+		einfo "Leaving pre-existing genkernel cache at ${ROOT}var/cache/genkernel/4.3.10 alone."
 	else
-		einfo "Copying genkernel cache into /var/cache/genkernel/."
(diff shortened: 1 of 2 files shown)


geaaru:
Two notes:
1. if you use `rsync` is better add `net-misc/rsync` on DEPEND
2. :thinking: i don't understand correctly why you need to copy/rsync files inside the `pkg_postinst` phase. Normally, ebuild/portage works in a sandbox that writes files under `/var/tmp/portage/<cat>/<pkg>/image/* and it's the portage that manages the files generated to the host filesystem. So, something like this:
```
dodir /var/cache/genkernel/
rsync -s ${S}/genkernel-cache/ ${D}/var/cache/genkernel/
```
Else you lose the relationship between files and the package that generates them.

I never used debian-sources kernel so i don't know well all the path. What files are generated inside `genkernel` directory?

cuantar:
1. I'll implement that. 
2. It's a cache that holds things `genkernel` has compiled, like `busybox` and `lvm`.  It should not be associated with the package, because in that case there are problems with file collisions when installing another version in parallel.  The files don't change when the kernel changes.  The sandbox gets in the way a bit, because the cache is inaccessible at the time the `initramfs` is being built.  To solve this problem, I added instructions to copy the cache into the workdir before building, and then to copy it back afterward in case it has changed or has just been created.

Maybe there is a better way to do this.

cuantar:
The cache is essentially throwaway and shouldn't be tracked by the package manager.

cuantar:
When we have Dracut support in the kernel ebuild, none of this will matter because we can replace Genkernel with it.

cuantar:
I think I've figured out what's going on here.

During the first time a package is built by `metro`, there is no `binpkg` already, so when the `genkernel` cache files are created, they are installed on the host.  During subsequent runs, the `binpkg` from the first run exists, so `metro` installs it; this install fails, because the `genkernel` cache files are copied from the `WORKDIR` and the `WORKDIR` is not included in the `binpkg`.

I'm investigating alternative avenues.

cuantar:
Yeah, the better way to do this is with `addwrite /var/cache/genkernel` on the line above the call to `genkernel` itself.

PR #87, opened 2024-08-24
[bugfix] [kernel] Fix `genkernel` cache and bump kernel to `debian-sources-6.10.4_p1`, following trixie

cuantar:
The `genkernel` cache in `metro` builds was not working after the first merge, because the way it was implemented, the ebuild copied the cache to the host's filesystem from `WORKDIR`; for `binpkgs`, the files aren't generated, so the directory in `WORKDIR` was never populated (and does not exist)!

This PR fixes that Bug by use of `addread`/`addwrite` to make `/var/cache/genkernel` in the host's filesystem accessible within the sandbox, at the time that the `initramfs` is created.  That way, the cache can be used if it exists or created if it doesn't; it won't be counted as part of the package (to avoid collisions if two versions are installed); and it won't interact with binary packages (as used by `metro`) at all.

Additionally, this PR bumps the kernel version to `debian-sources-6.10.4_p1` to follow Debian trixie; this version is compatible with `openzfs` per the `META` file in its repository.  The revision of `debian-sources-6.1.99_p1` is bumped to account for changes in the ebuild logic.  (If we skip a rev here, it's because there was a broken one in the middle that didn't escape local testing.)


geaaru:
Just a note. If you can it's better to avoid reference issues or pulls of your internal Gitea installation. It's few readable for users.
The comment contains a reference to issue 36 that is not related to this. Is of your Gitea server?
The right syntax is `macaroni-os/mark-issues` + `#` + `<num-issue>`

geaaru:
I think that you can rebase the branch that you generate this PR with:
```
$> git log -n3 
$> # read git hash of the third commit
$> git rebase -i <hash>
```
on textual editor:
```
r <first-hash> [bugfix] [kernel] ...
fixup <second-hash> [bugfix] ...
```
This will merge the two commits with a more simple commit. The first commit contains a command that is later removed. It doesn't make sense to have a history of the commit in this case. If you reference the PR the history will be here.

cuantar:
Yeah, that's a commit message from a merge commit on my internal server, referring to the local PR.  I didn't change any of them.

I suppose I can clean them up with `rebase`, because they are confusing when they end up in the upstream tree.

cuantar:
I see your second comment too, and understand from it that your preference for the upstream tree is to keep patches minimal, rather than attempting to track the development history as it happened.  Acknowledged!

geaaru:
Yeah, at the moment is the best way because I'm trying to keep things aligned between `master` and `mark-testing` branches with cherry-pick. When we have stabilized the mark release probably `next` branch will probably be updated less.

geaaru:
The merge branch is not usable, the two trees will be soon different and already have differences. So, to do a single cherry-pick is lovely more faster for me :).

cuantar:
Obsolete!

PR #90, opened 2024-08-25
[feature] [kernel] Autogen kernel ebuilds :D  (based on `master` branch)

cuantar:
Particular considerations:
- Track Debian unstable, testing, and stable branches
- Check compatibility with OpenZFS (except for `sid` :boom::sunglasses:)
- Move everything that differentiates ebuild-versions into the `autogen.yaml`
- Kernel source downloaded from kernel.org; Debian patches from debian.org and configs
- Removed keywords from `sid` kernel so that it is masked, for safety :safety_vest: 
- Added `mix-ins` for selecting kernel versions by branch
- Prevent kernel, initramfs, and modules from being removed upon package removal (or upgrade)
- Keep old kernel, initramfs, modules with `.old` suffix
- Ensure `/boot` is mounted during `pkg_postinst` phase; the original ebuild seems not to work in some edge cases

There's a `generator.py`, and a prototype `generator.sh` that demonstrates what's supposed to be happening.

This patch has been tested locally in a dev tree to ensure it builds and doesn't break regens.  Then it was pushed to the kitz0rsoft tree, where it was tested in `metro` builds (both initial and subsequent) and on several physical hosts with different subarchitectures and boot topologies.  Some hosts use `savedconfig` and others don't.

~@geaaru How's this for an easy PR?~  lol, perhaps not anymore...


cuantar:
Here it is, for when it's ready.  I want to test it in metro builds first.

cuantar:
TODO:  Mask `sid` so it doesn't get pulled in as an upgrade. It's in a different slot, but it's "special" and shouldn't supplant `trixie` as the default.

cuantar:
I didn't like how, with the slots like `trixie/6.10.4_p1`, the installed kernel files are all erased when the kernel is updated.  That could break someone's boots :hiking_boot::thinking:.  The best way does seem to be `6.10.4_p1/trixie`.  I'd hoped to be able to add `sys-kernel/debian-sources:trixie` (or `bookworm`) to `@world, and track only that release, without having to _e.g._ mask the `trixie` kernels on a system that wants to run `linux-6.1.xyz`.

cuantar:
Added `mix-ins` for selecting a kernel branch by name.  Ebuilds are generated for both kinds of slots, because I can see benefits from doing it this way.

cuantar:
Changed back to $branch/$PVR, and modified the ebuild to install the kernel, modules, and initramfs to temporary locations before moving them into place during `pkg_postinst`.  That way, the files needed for the running kernel and for booting won't be erased during a kernel upgrade.  The user can clean them up later at a convenient time.

cuantar:
The patch in the PR seems to be working well now.  Debian updated the `sid` kernel and a few of my machines that I'd set to that branch picked it up.  Will open this PR after the `trixie` kernel updates.

cuantar:
Closing on this branch, since it's been merged in `mark-testing`.

PR #95, opened 2024-09-01
[core-kit] sys-libs/minizip-ng - Drop old patch

geaaru:
Closes: macaroni-os/mark-issues#111


geaaru:
Tested in Macaroni [Binary](https://github.com/macaroni-os/macaroni-funtoo/commit/a9f137a3a6a4bc0a76e89525f34f6e307645e897)

PR #113, opened 2024-09-18
sci-mathematics/rstudio-bin

org-tekeli-borisp:
Closes: macaroni-os/mark-issues#126

```
 ╰ $ doit
INFO     Autogen: sci-mathematics/rstudio-bin (latest)                                                                                                                                                                                                                                                              
INFO     Download from https://download1.rstudio.org/electron/focal/amd64/rstudio-2024.04.2-764-amd64-debian.tar.gz verified as valid tar.gz archive.                                                                                                                                                               
INFO     Created: rstudio-bin-2024.04.2.764.ebuild   ```

### DIFF
--- a/science-kit/curated/sci-mathematics/rstudio-bin/autogen.py
+++ b/science-kit/curated/sci-mathematics/rstudio-bin/autogen.py
@@ -4,10 +4,8 @@ from bs4 import BeautifulSoup
 import re
 
 async def generate(hub, **pkginfo):
-	html_data = await hub.pkgtools.fetch.get_page(
-		f"https://www.rstudio.com/products/rstudio/download/"
-		)
-	soup = BeautifulSoup(html_data, "html.parser")
+	html_data = await hub.pkgtools.fetch.get_page(f"https://posit.co/download/rstudio-desktop/")
+	soup: BeautifulSoup = BeautifulSoup(html_data, "html.parser")
 	best_archive = None
 	for link in soup.find_all("a"):
 		href = link.get("href")


geaaru:
@org-tekeli-borisp have you tried to emerge this ebuild? I fear that with the change of the domain is changed also the way they supply the tarball.

From your `doit` we use `focal` release. Is this correct or does we need to use the electron (Debian 12) tarball?

geaaru:
I merge this in any case due fix the regen. Let me know if the ebuild is broken. Maybe we could keep the issue open if doesn't work.

org-tekeli-borisp:
@geaaru Indeed I had to test emerge too! I did it now and the emerge reports missing dependencies
image

org-tekeli-borisp:
Stop! Tested before ego sync!

org-tekeli-borisp:
After ego sync the same picture. So I assume I didn't broke the ebuild and the new downlaod url leads to the same tar.gz as the previous one
image

org-tekeli-borisp:
So we solved the regen issue, but the ebuild itself still "broken"

org-tekeli-borisp:
@geaaru  This package sci-mathematics/rstudio-bin does have dependencies on packages we don't have in the tree anymore- e.g. java stuff like dev-java/aopalliance,  dev-java/gin,  dev-java/javax-inject, dev-java/validation-api.

org-tekeli-borisp:
Removing them lets the emerging to start
image
but it fails on R
image

org-tekeli-borisp:
@geaaru Seems like this package needs a bigger refactoring or maybe a removing from tree?

geaaru:
Well, I agree on removing `dev-java/aopalliance` from dependency, considering that is a `*-bin` probably doesn't need any other dependencies.
Instead it's better to open an issue for `dev-lang/R` that is broken. But it seems something related to the LAPACK staff. What your lapack implementation is configured?

PR #138, opened 2024-10-09
[dev-kit] dev-util/cunit add patches

geaaru:
Closes: macaroni-os/mark-issues#154


org-tekeli-borisp:
@geaaru what about multlib? should we remove this support too?

geaaru:
yeah, i saw. But it seems that to compile it copy sources to a different directory and atm I haven't time to refactor completely this package. Probably we need a specific task to replace multilib globally but it's better focus on more critical tasks. Unluckily, we are few.

PR #158, opened 2024-10-22
[feature] [kernel] Autogen kernel ebuilds :D

cuantar:
Particular considerations:
- Track Debian unstable, testing, and stable branches
- Check compatibility with OpenZFS (except for `sid` :boom::sunglasses:)
- Move everything that differentiates ebuild-versions into the `autogen.yaml`
- Kernel source downloaded from kernel.org; Debian patches from debian.org and configs
- Removed keywords from `sid` kernel so that it is masked, for safety :safety_vest: 
- Added `mix-ins` for selecting kernel versions by branch
- Prevent kernel, initramfs, and modules from being removed upon package removal (or upgrade)
- Keep old kernel, initramfs, modules with `.old` suffix
- Ensure `/boot` is mounted during `pkg_postinst` phase; the original ebuild seems not to work in some edge cases
- Changed from using https://packages.debian.org to one of the `.deb` package repo mirrors as the source for data about the current version

There's a `generator.py`, and a prototype `generator.sh` that demonstrates what's supposed to be happening.

This patch has been tested locally in a dev tree to ensure it builds and doesn't break regens.  Then it was pushed to the kitz0rsoft tree, where it was tested in `metro` builds (both initial and subsequent) and on several physical hosts with different subarchitectures and boot topologies.  Some hosts use `savedconfig` and others don't.

~@geaaru How's this for an easy PR?~  lol, perhaps not anymore...

New in this PR: Rebased against the `mark-testing` branch; will leave the old one open until this one is merged.


geaaru:
Good to have this in our `mark-testing` branch and soon integrate a testing tarball that prepare the kernel too.
Thanks

geaaru:
Eventually in the near future we can prepare a `kernel-kit` instead of using `core-kit`.

PR #159, opened 2024-10-23
[core-kit] app-emulation/incus review autogen

geaaru:
Closes: macaroni-os/mark-issues#172


cuantar (approved):
This all seems reasonable to me.

PR #166, opened 2024-10-26
[core-kit] Drop old GCC ebuilds

geaaru:
Closes: macaroni-os/mark-issues#77


cuantar (approved):
Wow!  Most of these are hidden in such a way that they can't be emerged:

```
$ ls /var/git/meta-repo/kits/*/sys-devel/gcc
files/  Manifest  gcc-11.3.0-r1.ebuild  gcc-12.2.0.ebuild  gcc-12.3.0.ebuild  README.txt
```
Removing them likely is harmless at this point.  I agree, do it.

PR #175, opened 2024-10-27
[desktop-kit] [bugfix] Remove failing patches from `app-crypt/xca` ('mark-testing')

cuantar:
After removing the patches, `xca` successfully installs again.


cuantar:
Fixes macaroni-os/mark-issues#158.

PR #186, opened 2024-11-02
[core-kit] [bugfix] Fix genkernel compatibility with updated linux-headers

cuantar:
Since `linux-6.8`, changes to the kernel headers necessitate a patch for `busybox-1.36.1` and `busybox-1.37.0`.

This PR does the following:
- Add patch from `sys-apps/busybox` to genkernel's installed files
- (genkernel automatically applies the patch)
- Bump genkernel's internal busybox version to match the current one in MARK tree

Tested with `metro` from a dev tree.  Originally found the issue during `metro` builds on `mark-testing`.  In the `metro` environment, there aren't any cached genkernel `initramfs` segments, so `busybox` needs to be built.  With this patch, `metro` builds succeed on the `mark-testing` tree.

### DIFF
--- a/core-kit/curated/sys-kernel/genkernel/autogen.yaml
+++ b/core-kit/curated/sys-kernel/genkernel/autogen.yaml
@@ -8,7 +8,7 @@ genkernel_rule:
         version: 4.3.10
         # so that people get the dmraid gz -> bz2 fix
         revision:
-          '4.3.10': '3'
+          '4.3.10': '4'
         patches:
           - 4.3.10-fix-modinfo-error.patch
         github:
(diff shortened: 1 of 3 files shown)


geaaru:
It's ok for me. Just a note: why we can just using the busybox version from the tree or at least using patches from there? This avoid to add patches in two different places. Do you think that could be an issue?

Maybe, you can get patches from the tree directly like we do for nvidia kernel modules for the sources in the `src_unpack` phase. Never tested, so, i don't know if this could be correct.

geaaru (approved):

cuantar:
> It's ok for me. Just a note: why we can just using the busybox version from the tree or at least using patches from there? This avoid to add patches in two different places. Do you think that could be an issue?
> 
> Maybe, you can get patches from the tree directly like we do for nvidia kernel modules for the sources in the `src_unpack` phase. Never tested, so, i don't know if this could be correct.

We are using the busybox version from the tree, and the patch comes directly from there.  It's an interesting idea to copy the patch in the ebuild from the tree at `/var/git/meta-repo` instead of `FILESDIR` for this package, since the patch is the same. I think I'll do that.

cuantar:
Well, except for one problem:  `busybox` in the tree might change, but genkernel builds its own that is decoupled.  Perhaps it makes more sense to leave genkernel alone once it is working, because I don't really intend for it to be a long-term solution.  It was just a way to get things stable so I could implement Dracut.

geaaru:
Well, busybox could be updated but in the autogen branches. The stable branches will be managed with a cd/ci testing chain so this could be not an issue in the near future.

PR #192, opened 2024-11-08
[desktop-kit] [bugfix] Update `passwordsafe` autogen.py: `name` -> `body`

cuantar:
The string "Linux" seems to occur more often in the `body` field than in the `name`, where it is never found (anymore?).  In fact, `name` is most often empty, or only a numerical version.  This patch fixes the autogen by pointing the search at `body` instead of `name`.


org-tekeli-borisp:
I would like into cherry-peak this into master as passwordsafe breaks tree regeneration on master too

org-tekeli-borisp:
done!

PR #219, opened 2024-11-25
Update bashlogin-banner

coffnix:
new script migrated to macaroni OS


geaaru:
@coffnix please, rename the package with `sys-apps/live-vipnix-scripts` so I can close this PR

PR #226, opened 2024-12-02
Rebranding www-client/firefox and mail-client/thunderbird

org-tekeli-borisp:
relevant to 
- https://github.com/macaroni-os/mark-issues/issues/225
- https://github.com/macaroni-os/mark-issues/issues/224


cuantar:
Great idea to do things like this.  It's quick and easy, and user-visible.

PR #256, opened 2025-01-17
[core-kit] sys-fs/zfs Update ZFS autogen.yaml to remove version matching rules

org-tekeli-borisp:
Summary
========
* Removed specific version matching patterns and commented-out sections in the ZFS autogen.yaml configuration. This streamlines the file and relies on default behavior for release handling.

Test Plan
========
* Doit
```
 ╰ $ doit
INFO     Autogen: sys-fs/zfs (latest)                                                                                                                                                                                        
INFO     Autogen: sys-fs/zfs-kmod (latest)                                                                                                                                                                                   
INFO     Created: ../zfs-kmod/zfs-kmod-2.3.0.ebuild                                                                                                                                                                          
INFO     Created: zfs-2.3.0.ebuild   
```
* Make it available in a own repo tree or by seeding into a local overlay
* Emerge it
* Running it
```
~ # zfs version
zfs-2.3.0-r0-funtoo
zfs-kmod-2.3.0-r0-funtoo
```

### DIFF
--- a/core-kit/curated/sys-fs/zfs/autogen.yaml
+++ b/core-kit/curated/sys-fs/zfs/autogen.yaml
@@ -6,13 +6,6 @@ zfs_and_friends:
       user: openzfs
       repo: zfs
       query: releases
-      match: zfs-(2\.2\.[0-9].*)
-#      include:
-#        - prerelease
-#      transform:
-#        - kind: string
-#          match: '-rc'
-#          replace: '_rc'
   packages:
     - zfs
     - zfs-kmod


core-kit/curated/sys-fs/zfs/autogen.yaml
```diff
@@ -6,13 +6,6 @@ zfs_and_friends:
       user: openzfs
       repo: zfs
       query: releases
-      match: zfs-(2\.2\.[0-9].*)
```
geaaru, line 9: I think that this could be:
```yaml
match: zfs-(2\.[0-9]\.[0.9].*)
```

core-kit/curated/sys-fs/zfs/autogen.yaml
```diff
@@ -6,13 +6,6 @@ zfs_and_friends:
...
-#      include:
-#        - prerelease
-#      transform:
-#        - kind: string
-#          match: '-rc'
-#          replace: '_rc'
```
geaaru, line 15: We can maintain the transform in order to avoid having broken ebuild with new release candidates and not gentoo like version string.

geaaru (changes requested):

geaaru:
Maybe you can change `-funtoo` with `-macaroni` for suffix driver name ;)

core-kit/curated/sys-fs/zfs/autogen.yaml
org-tekeli-borisp: @geaaru  Maybe we should use the following to prevent the rc's an get only finalized releases?
> match: zfs-([0-9]\.[0-9]\.[0.9])

geaaru: Correct.

Well, this will be more easy with the new autogen tool. In the release Github metadata there are flag about prerelease/draft that atm i just excluded, see:

https://github.com/macaroni-os/mark-devkit/blob/master/pkg/autogen/generators/builtin_github.go#L324

So, hoping that community use these flags the problem will be less critical.
But it could add the support of the `exclude` on github generator too and skiping `*rc` stuff like I did for `dirlisting`.
My note was related to avoid adding always extra effort on update regex/rules with the need to follow the upstream releases. But, considering that we need to migrate to new tools we can leave more controlled the rules here and choice the right way in the new branch.

core-kit/curated/sys-fs/zfs/autogen.yaml
geaaru: Like described in the other section we keep stable release here.

PR #277, opened 2025-02-02
[core-kit] genkernel autogen.yaml: update source for boost tarball

cuantar:
This small change fixes the genkernel autogen.


cuantar:
Closes macaroni-os/mark-issues#295 .

geaaru:
@cuantar the `Closes` tag is for git commit ;)

cuantar:
Why didn't the Issue get closed?  I thought `Closes` refers to Issues in another repo that way.

geaaru:
Yeah, correct. The `Closes` reference the issue but to trigger the close you need to add the string in the git commit comment.

This normally works. With the introduction of the sub issues I think that this does not always work correctly. It seems like a GitHub issue.

PR #279, opened 2025-02-02
[core-kit] Add intel64 subarches alderlake and rocketlake

cuantar:
Hooray, a feature!

I've been using these locally for a few months now.  Here are subarches for `rocketlake` (Intel 11th gen) and `alderlake` (12th-gen).  They seem to do what's expected of them and I haven't detected any monkey business so far.


geaaru:
Can you create a PR for the `mark-xl` branch too? I don't think that in this case we can do a cherry pick.

geaaru:
Thanks for this enhancement!

PR #290, opened 2025-02-25
[core-kit] Update `net-misc/openssh` version pin to `9.9_p2`

cuantar:
Peripherally:  I think perhaps we should move `net-misc/openssh` into `core-server-kit`.


cuantar:
Performed automated tested in a fresh `stage3` container, from `doit` through `emerge` from an overlay; then verified that after the `sshd` service restarted, the container was still accessible via `ssh`, and could get out to a remote machine.

geaaru (approved):

PR #295, opened 2025-03-10
Qt6 inital commit

linuxexplorer:
Added for testing:
/eclass/qt6-build.eclass
/qt-kit/autogen.kit.d/templates
/qt-kit/autogen.kit.d/qt6spec.yml
/qt-kit/merge.kit.d/qt6.yml

Known issues:
qtmultimedia does not successfully merge with the gstreamer use flag

### DIFF
--- a/eclass/qt6-build.eclass
+++ b/eclass/qt6-build.eclass
@@ -1,0 +1,180 @@
+# Copyright 2021-2023 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+# @ECLASS: qt6-build.eclass
+# @MAINTAINER:
+# qt@gentoo.org
+# @SUPPORTED_EAPIS: 8
+# @BLURB: Eclass for Qt6 split ebuilds.
+# @DESCRIPTION:
+# This eclass contains various functions that are used when building Qt6.
+# Requires EAPI 8.
+
+case ${EAPI} in
+	7) ;;
+	*) die "${ECLASS}: EAPI ${EAPI:-0} not supported" ;;
+esac
+
+if [[ -z ${_QT6_BUILD_ECLASS} ]]; then
+_QT6_BUILD_ECLASS=1
+
+[[ ${CATEGORY} != dev-qt ]] &&
+	die "${ECLASS} is only to be used for building Qt 6"
+
+# @ECLASS_VARIABLE: QT6_MODULE
+# @PRE_INHERIT
+# @DESCRIPTION:
+# The upstream name of the module this package belongs to. Used for
+# SRC_URI and EGIT_REPO_URI. Must be set before inheriting the eclass.
+: ${QT6_MODULE:=${PN}}
+
+# @ECLASS_VARIABLE: VIRTUALX_REQUIRED
+# @DESCRIPTION:
+# For proper description see virtualx.eclass man page.
+# Here we redefine default value to be manual, if your package needs virtualx
+# for tests you should proceed with setting VIRTUALX_REQUIRED=test.
+: ${VIRTUALX_REQUIRED:=manual}
+
+inherit cmake virtualx
+
+HOMEPAGE="https://www.qt.io/"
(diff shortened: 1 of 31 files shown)


eclass/qt6-build.eclass
```diff
@@ -0,0 +1,180 @@
...
+# Build directory for out-of-source builds.
+: ${QT6_BUILD_DIR:=${S}_build}
+
+IUSE="debug test"
+
+if [[ ${QT6_BUILD_TYPE} == release ]]; then
+	RESTRICT="test" # bug 457182
+else
+	RESTRICT="!test? ( test )"
+fi
```
geaaru, line 93: Hi @linuxexplorer, IMHO, I think that we can begin to drop all Gentoo stuff no more needed when we use a generation pattern based on templates. The Gentoo way was, we create an eclass to reduce effort and reduce duplication in the ebuild. With the generation of the ebuild based on templates this is no more a real requirement.
The SLOT, the SRC_URI and other options are automatically generated by `mark-devkit` and we can just import them in the template.

qt-kit/autogen.kit.d/qt6spec.yml
```diff
@@ -0,0 +1,288 @@
+# file qt6spec.yml
+dev-qt:
+  generator: builtin-dirlisting
+  template:
+    engine: helm
+
+  defaults:
+    category: dev-qt
```
geaaru, line 8: You can use the selector in order to automatically get new releases automatically. This is the reason because we work on `mark-unstable` before the rest. The filter of the version could be done in the testing/stable release:
```
defaults:
   vars:
      release: 6.8
      slot: 6
   selector:
       - '<6.9'
   assets:
       - name: "{{ .Values.pn }}-everywhere-src-{{ .Values.version}}.tar.xz"
         prefix: https://download.qt.io/archive/qt/{{ .Values.release }}/{{ .Values.version }}/submodules/
```

The `assets` using render engine on *prefix* could be defined in the default section only one time :champagne:

qt-kit/autogen.kit.d/qt6spec.yml
```diff
@@ -0,0 +1,288 @@
...
+  packages:
+    - qt3d:
+        vars:
+          desc: "Qt 3D provides a fully configurable renderer that enables developers to quickly implement any rendering pipeline that they need"
+          versions:
```
geaaru: I think that you can avoid to set `versions` var and left to `builtind-dirlisting` to catch last available version adding eventually a filter if it takes wrong links.

qt-kit/merge.kit.d/qt6.yml
```diff
@@ -0,0 +1,53 @@
+# cat qt6.yml
+sources:
+# We need this to inherit the eclass used for kit cache JSON generation
+- name: "qt-kit"
```
geaaru: You need always the `core-kit` with eclasses to correctly generate the reposcan files and the elaborated ebuild with all elaborated variables.
You don't need instead the `qt-kit` as source.

qt-kit/autogen.kit.d/templates/qt3d.tmpl
```diff
@@ -0,0 +1,31 @@
+# Distributed under the terms of the GNU General Public License v2
```
geaaru, line 1: For simple package we can use the common template instead like I do for llvm-kit:
https://github.com/macaroni-os/kit-fixups/blob/mark-unstable/llvm-kit/autogen.kit.d/base.yml#L7
and set *rdpend*, *depend* and *body* in this way:
```
   - qt3d:
         vars:
            rdepend: |
               app/foo
            depend: |
                app/foo2
            body: |
                src_configure() {
	           local mycmakeargs=(
	       	-DQT_FEATURE_qt3d_system_assimp=ON
                   )

	           qt6-build_src_configure
               }
```
And so remove a lot of templates files. But it's also fine in this way if you prefer.
Consider that on rdepend, depend, bdepend and body you can use the render engine too for example for SLOT ( {{ .Values.slot }})

PR #305, opened 2025-03-20
[python-modules-kit] Add `du_pep517: generator` to `yq` YAML block

cuantar:
This patch fixes `yq` emerging.


cuantar:
Closes: macaroni-os/mark-issues#322

PR #324, opened 2025-03-22
[core-gl-kit] [media-kit] Autogening fonts

org-tekeli-borisp:


geaaru (approved):
It seems ok for me. @cuantar can you review this? Considering that this PR introduces a lot of packages it's better to review by different devs, IMHO

geaaru:
Move things ahead for testing these packages with mark-testing and mark-unstable

PR #328, opened 2025-03-22
[core-kit] sys-apps/shadow Autogening

org-tekeli-borisp:


geaaru:
The package `sys-apps/shadow` is been autogen with mark-devkit.
See https://github.com/macaroni-os/kit-fixups/blob/mark-unstable/core-kit/autogen.kit.d/core.yml#L544

I dropped all patches because enabling subuid/subgid seems something used with LXD and container. Do you see any cons about this?

Why you drop the cracklib depedency? Is no more needed the check of the dictionary?

org-tekeli-borisp:
> Why you drop the cracklib depedency? Is no more needed the check of the dictionary?

Because of failure in the configuration phase as they were reported as unknown parameters.

geaaru:
> > Why you drop the cracklib depedency? Is no more needed the check of the dictionary?
> > Because of failure in the configuration phase as they were reported as unknown parameters.

But could be this something yet valid if we have cracklib deps installed ? https://github.com/macaroni-os/kit-fixups/blob/mark-unstable/core-kit/autogen.kit.d/templates/shadow.tmpl#L132

org-tekeli-borisp:
> But could be this something yet valid if we have cracklib deps installed ? https://github.com/macaroni-os/kit-fixups/blob/mark-unstable/core-kit/autogen.kit.d/templates/shadow.tmpl#L132

Seems like shadow is autogened  in mark-unstable already?

geaaru:
Yeah, exactly. But i just figured out now that the tarball is changed. Maybe the tag is been re-tagged. I will check tomorrow.

PR #331, opened 2025-03-22
[core-kit] sys-apps/baselayout Add autogen support for sys-apps/basel…

org-tekeli-borisp:
…ayout

Introduces an autogen.yaml to automate baselayout package generation. Replaces static ebuild with a template-based approach using package metadata. Removes obsolete audit and Manifest files.


geaaru:
@org-tekeli-borisp i don't agree with this. It doesn't make sense continue to follow Gentoo for this.

I already totally rewritten the ebuild and it uses direct files from kit-fixups.

Have you see this:
https://github.com/macaroni-os/kit-fixups/blob/mark-testing/core-kit/curated/sys-apps/baselayout/baselayout-3.0.ebuild

I think that is better for your tree using kit-fixups mark as start point instead of the old next tree. Could be this the cause of your PR?

org-tekeli-borisp:
Roger!

PR #354, opened 2025-05-03
[core-server-kit] change "Funtoo Linux" -> "Macaroni OS" and replace logo in `www-servers/nginx` example html file

cuantar:
Closes: macaroni-os/mark-issues#330


geaaru:
Move ahead this and thanks. Maybe you can autogen this with mark-devkit as next step ;)

PR #355, opened 2025-05-03
[python-modules-kit] add `importlib_metadata` to deps for `setuptools_scm`

cuantar:
Closes: macaroni-os/mark-issues#355


geaaru:
@cuantar i asked to @org-tekeli-borisp to open this because add `importlib_metadata` add a deps cycle that must be managed correctly in another way. This need more attention.

PR #356, opened 2025-05-05
[core-kit] Use generalized, complete version of more-uarches-patch for `debian-sources`

cuantar:
Previous PR included the `lite` version of the patch, that only introduces the generic optimization levels.  This one is the generalized version for all 6.1.x+ kernels that adds the subarches.


cuantar:
Closes: macaroni-os/mark-issues#354.

PR #359, opened 2025-05-13
[mark-xl] Add a few more packages to `mark-xl`

cuantar:
core-server-kit:
  - app-admin/ansible-core
  - app-emulation/libvirt

net-kit:
  - net-fs/autofs

security-kit:
  - app-admin/sshguard

### DIFF
--- a/core-server-kit/merge.kit.d/tools.yml
+++ b/core-server-kit/merge.kit.d/tools.yml
@@ -34,6 +34,7 @@ target:
   atoms:
     - pkg: app-admin/ansible
     - pkg: app-admin/ansible-base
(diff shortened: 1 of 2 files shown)


cuantar:
Actually, I think `autofs` would be better in `core-server-kit`, because it's not only concerned with mounting network shares.

geaaru:
As shared in Discord the package `ansible-core` could be not needed. FWIS seems that `ansible-core` is for Python >=3.11 and `ansible-base` is the stack for old versions of python >2.7.
I will move ahead this as-is to study and test this in the tree.

PR #360, opened 2025-05-13
[mark-xl] Add `x11-proto/xwaylandproto` to `xorg-kit`

cuantar:
Closes: macaroni-os/mark-issues#367


cuantar:
Probably not needed; closing.

PR #361, opened 2025-05-13
[core-kit/portage-kit] add correct `PORTDIR_CACHE_METHOD directive` to `/etc/eixrc/...`

cuantar:
We need to add 
```
PORTDIR_CACHE_METHOD="parse#metadata-md5#metadata-flat#assign"
```
to the default `eixrc` in order for `eix` to work properly with `core-kit` (and more?) in `mark-xl`.

Here's a trivial patch to add the necessary line in the right place.


cuantar:
There, it's in the correct place and seems to work as expected:
```
cat /etc/eixrc/mark.conf 
PORTDIR_CACHE_METHOD="parse#metadata-md5#metadata-flat#assign"
```

cuantar:
Closes: macaroni-os/mark-issues#331

PR #367, opened 2025-06-07
[core-kit] Update pin to `openssl-1.1.1w` and remove failing patch

cuantar:
This change is pretty small.  The removed patch added a header that was needed for compilation to succeed previously, but is no longer necessary.


cuantar:
Tested locally with the original `metro` by building a `stage3` with the new `openssl` version.

cuantar:
Closes: macaroni-os/mark-issues#89

geaaru:
Closing this because this package is now autogen on mark-unstable

PR #371, opened 2025-06-30
[mark-xl] Rebase from `mark-testing` -> `mark-unstable`

cuantar:
We've been pulling packages directly from `mark-testing` into `mark-xl`.  We should not do that, and instead pull from `mark-unstable`.  This PR breaks the direct link between `mark-xl` and `mark-testing`.

### DIFF
--- a/core-hw-kit/merge.kit.d/base.yml
+++ b/core-hw-kit/merge.kit.d/base.yml
@@ -2,11 +2,11 @@ sources:
 # We need this to inherit the eclass used for kit cache JSON generation
 - name: "core-kit"
   url: "https://github.com/macaroni-os/core-kit"
-  branch: "mark-testing"
+  branch: "mark-unstable"
 
 - name: "core-hw-kit"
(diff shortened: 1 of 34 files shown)


geaaru:
@cuantar I think that we can do also cleanup. The merge phase from mark-testing requires multiple sources. Using mark-unstable this could be simplify. IIRC the mapping is pretty 1:1 or with very little differences between the branches.

I mean, whilte now with `repoA` you have `repoA`+`repoB`+`repoC`+`repoD`, now probably it's sufficient having `repoA` and `repoB` as source.

This will speedup the mark-devkit task because will reduce the number of repositories to convert in JSON for the analysis.

Let me know if you prefer move ahead things as-is and later cleanup the merge specs.

cuantar:
> Let me know if you prefer move ahead things as-is and later cleanup the merge specs.

Good idea to do it now. I'll include the cleanup in this PR.

geaaru (approved):

PR #372, opened 2025-06-30
[net-kit] Add mail-client/alpine-2.26 autogen

cuantar:
I've tested the resulting ebuild locally.  It builds, and the installed files are as expected.  So far I've encountered no issues with the `alpine` binary.

### DIFF
--- a/net-kit/autogen.kit.d/base.yml
+++ b/net-kit/autogen.kit.d/base.yml
@@ -315,6 +315,85 @@ flatpak:
             }
 
 
+alpine:
+  generator: builtin-noop
+  template:
+    engine: helm
+
+  defaults:
+    category: mail-client
+    files_dir: "{{ .Values.pn }}"
+
+  packages:
+    - alpine:
+        template: ../../templates/simple.tmpl
+        assets:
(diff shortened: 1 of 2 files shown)


net-kit/autogen.kit.d/base.yml
```diff
@@ -315,6 +315,85 @@ flatpak:
...
+                myconf+=( --with-interactive-spellcheck=/usr/bin/hunspell )
+              elif has_version "app-text/aspell"; then
+                myconf+=( --with-interactive-spellcheck=/usr/bin/aspell )
+              fi
+
+              if use ssl; then
+                myconf+=(
+                  --with-ssl-include-dir="${EPREFIX}"/usr/include/openssl
+                  --with-ssl-lib-dir="${EPREFIX}"/usr/$(get_libdir)
+                  --with-ssl-certs-dir="${EPREFIX}"/etc/ssl/certs
```
geaaru, line 382: I think that at some point we can stop to use EPREFIX variable. Gentoo custom chroot it doesn't make sense anymore having fchroot and/or another tecnology.
Btw, moving ahead things

geaaru (approved):

cuantar:
Understood!  I'll leave that variable out for future packages.

PR #373, opened 2025-07-01
[kernel-kit] Autogen sys-kernel/debian-sources and use Dracut for initramfs generation.

cuantar:
Doing some more testing before marking it ready to merge.

Still TODO is to look at the `META` file for the `openzfs` tag that corresponds to the version in our tree, in order to get the kernel min and max supported versions.


kernel-kit/autogen.kit.d/sys-kernel/debian-sources/files/config-extract-6.6
```diff
@@ -0,0 +1,229 @@
...
+			features.append(flav)
+
+PROG = "config-extract"
+
+
+def usage():
+	print(
+		"""This work is free software.
+
+Copyright 2011-2024 Funtoo Solutions. You can redistribute and/or modify it under
```
geaaru, line 75: Maybe this could be updated :)

cuantar: I'm not sure about that.  I haven't touched that file, so it's still a Funtoo original.  To me, it seems wrong to remove the original author's comment like that if the file hasn't been changed.

cuantar:
I think this is good enough to merge.   But:  The ebuild template is heavily modified from the Funtoo version, with much of the original structure intact, and should probably be replaced with cleaner code at some point.

cuantar:
I'm also not using the recent updates to the custom generator, so the generator script could be improved.

geaaru:
Having a start point is an important result. The software like the autogen can be always improved :) . A software that doesn't improve and evolve it's just died. Having a `mark-unstable` branch is also means having a place with test new things without broken the stable trees.

Like you said one time, if you search perfection you will die before find it :) or with words similar. Better starts with something and later evolve.

cuantar:
Tested on two systems.  I think it's good to merge now.

PR #374, opened 2025-07-01
[net-kit] Fix redundant `autofs` definitions

cuantar:
One of them had the incorrect category, so it had no effect and we didn't notice.


geaaru:
`mark-devkit kit merge` doesn't drop a package. In order to complete this task you need create a PR to `net-kit` that drop the autofs package with the dropped category.

PR #375, opened 2025-07-01
[mark-iii] Add packages needed in `mark-iii`.

cuantar:
This PR is a work in progress.

### DIFF
--- a/desktop-kit/merge.kit.d/base.yml
+++ b/desktop-kit/merge.kit.d/base.yml
@@ -69,6 +69,10 @@ target:
 
     - pkg: dev-libs/libdbusmenu
 
+    - pkg: net-im/discord-bin
+    - pkg: net-im/slack-bin
+    - pkg: net-im/zoom-bin
+
     # Gkrellm stuff
     - pkg: app-admin/gkrellm
     - pkg: x11-plugins/gkrellm-bluez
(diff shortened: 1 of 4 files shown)


desktop-kit/merge.kit.d/base.yml
```diff
@@ -69,6 +69,10 @@ target:
 
     - pkg: dev-libs/libdbusmenu
 
+    - pkg: net-im/discord-bin
```
geaaru, line 72: https://github.com/macaroni-os/kit-fixups/blob/mark-unstable/desktop-kit/merge.kit.d/base.yml#L102

it seems already present (same for slack-bin)

cuantar: Ah, agreed.  Then, I need to autogen them in `mark-unstable` so that the update cadence is faster.  Discord in particular is frustrating, in that sometimes not applying updates will break voice calls.

geaaru: Discord, slack and Telegram are now autogen ;)

cuantar:
This PR still has my attention.  It will end up being a small number of existing packages in `mark-unstable` that need to be added to `mark-iii`.  Some packages ended up pulled into `mark-xl` from `mark-testing` without first going through `mark-unstable`; so those need to be autogenned, since we don't have ebuilds already except in `mark-testing`.

desktop-kit/merge.kit.d/base.yml
cuantar: thanks for doing that :D

cuantar:
All of the packages already in this PR are added now, except for games.  Games have another PR to autogen most of them in `mark-unstable`, so I'll close this one now.

PR #381, opened 2025-07-20
[core-kit] Remove consolekit from profiles

### LINKED ISSUE #366: [core-kit] Remove consolekit from existing profiles
`consolekit` is something obsolete. I read in the past about `consolekit2` but it seems less maintained.

I think that we can drop it from the existing `mix-in`/`profiles`.

geaaru:
Closes: macaroni-os/mark-issues#366


linuxexplorer:
I have mirrored the changes locally and everything appears ok on emerge
This is on an existing mark-iii kde-plasma-5 install

PR #384, opened 2025-08-03
[qt-kit] Minor corrections to dev-qt/qt{charts,positioning,webengine}

cuantar:
These three packages don't install without fixes.  The result is an inconsistent system with a mixture of old and new versions of Qt libraries.  All the Qt packages emerge at the expected version (now, `5.15.16`) with these changes.


geaaru:
On Phoenix all works because I use `--nodeps` and `anise-portage-converter` as resolver. Thanks for the fix. I will try to propagate the fix soon.

PR #387, opened 2025-09-08
[mark-unstable] Add `USE=zstd` for `sys-apps/kmod` to core flavor

cuantar:
This adjustment is needed for compressed module support.


coffnix (commented):
required for all architectures and flavors

PR #388, opened 2025-09-10
[kernel-kit] Update debian-sources for new Debian release; bugfixes

cuantar:
Specifically:
- add forky as new testing release
- change trixie to stable
- change bookwork to oldstable
- set KEYWORDS="*" for all versions

Bugfixes:
- dmraid -> mdraid, for mdadm
- send strictly "space-separated lists" to dracut
- `USE=mdadm` should pull in `sys-fs/mdadm` as dep
- ebuild needs to depend on `whip-catalog`

It will be necessary to mask sid in the profiles of testing/stable branches.  (Done)

### DIFF
--- a/kernel-kit/autogen.kit.d/debian.yml
+++ b/kernel-kit/autogen.kit.d/debian.yml
@@ -24,6 +24,7 @@ debian_sources:
       rdepend: |
         sys-apps/gawk
         dracut? (
+          sys-apps/whip-catalog
           sys-kernel/dracut
           virtual/dracut-mark
         )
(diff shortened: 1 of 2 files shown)


cuantar:
Hm, will include the KEYWORDS fix too.

cuantar:
Draft until I figure out why `dracut` is still failing to add the requested modules to the `initramfs` when called by the ebuild, but not when the same command is repeated manually.

cuantar:
Dracut expects and requires that arguments are passed as, strictly, "a space-separated list"; otherwise, the modules in the list aren't parsed correctly and it can't find any of them.  I've fixed it also in this PR by employing `xargs`.

cuantar:
Fixed!

```
$ lsinitrd  /boot/initramfs-debian-x86_64-6.12.43-debian1-mark
...
Arguments:  -N -f --xz -v --kmoddir '/lib/modules/6.12.43-debian1-mark' -a 'lvm crypt' --add-drivers 'dm-crypt'
...
```

geaaru (approved):

PR #393, opened 2025-12-31
Autogens to support kde6

linuxexplorer:
Bugs: macaroni-os/mark-issues#537
Note: switheroo-control and power-profiles-daemon are version-restricted due to requiring a newer glib for more recent versions.
All new ebuilds have been tested locally and merge.

### DIFF
--- a/gnome-kit/autogen.kit.d/base.yml
+++ b/gnome-kit/autogen.kit.d/base.yml
@@ -497,6 +497,176 @@ gnome_release_pkgs:
 
   packages:
 
+    - yelp:
+        vars:
+          inherit:
+            - gnome3
+            - meson
+          license: GPL-+2+
+          rdepend: |
+            dev-libs/glib:2
+            x11-libs/gtk:4
+            x11-libs/libadwaita
+            dev-libs/libxml2
+            dev-libs/libxslt
+            dev-db/sqlite:3
+            net-libs/webkit-gtk:6
+            >=gnome-extra/yelp-xsl-42.3
+            app-arch/xz-utils:=
+            app-arch/bzip2:=
+          depend: '${RDEPEND}'
+          bdepend: |
+            dev-util/itstool
+            sys-devel/gettext
+            virtual/pkgconfig
+          body: |
+            src_configure() {
+              local emesonargs=(
+                -Dlzma=enabled
+                -Dbzip2=enabled
+              )
+              meson_src_configure
+            }
+            pkg_postinst() {
+              gnome3_pkg_postinst
+            }
+            pkg_postrm() {
+              gnome3_pkg_postrm
+            }
+
+    - gnome-remote-desktop:
+        category: net-misc
+        vars:
+          inherit:
+            - gnome3
+            - meson
+            - systemd
+            - tmpfiles
+          license: GPL-2+
+          iuse: doc +rdp systemd +vnc
+          required_use: '|| ( rdp vnc )'
+          rdepend: |
+            x11-libs/cairo
+            x11-libs/libdrm
+            media-libs/libepoxy
+            dev-libs/glib:2
+            dev-libs/libei
+            x11-libs/libnotify
+            app-crypt/libsecret
+            media-video/pipewire:=
+            app-crypt/tpm2-tss:=
+            rdp? (
+              media-libs/nv-codec-headers
+              net-misc/freerdp:=[server]
+              sys-fs/fuse:3=
+              media-libs/libva
+              sys-auth/polkit
+              media-libs/shaderc
+              dev-uti/spirv-tools
+              x11-libs/libxkbcommon
+              media-libs/fdk-acc:=
+            )
+            vnc? (
+              net-libs/libvncserver
+            )
+            x11-wm/mutter[screencast]
+          depend: |
+            ${RDEPEND}
+            rdp? (
+              dev-util/vulkan-headers
+            )
+          bdepend: |
+            dev-util/gdbus-codegen
+            virtual/pkgconfig
+            doc? (
+              app-text/asciidoc
+            )
+          body: |
+            src_configure() {
+              local emesonargs=(
+                $(meson_use doc man)
+                $(meson_use rdp)
+                $(meson_use vnc)
+                $(meson_use systemd)
+                -Dsystemd_user_unit_dir="$(systemd_get_userunitdir)"
+                -Dtests=false  # Tests run xvfb-run directly
+              )
+              meson_src_configure
+            }
+            pkg_postinst() {
+              tmpfiles_process "${PN}-tmpfiles.conf"
+              gnome3_pkg_postinst
+            }
+            pkg_postrm() {
+              gnome3_pkg_postrm
+            }
+
+    - gnome-boxes:
+        vars:
+          license: LGPL-2+ CC-BY-2.0
+          desc: Simple GNOME application to access remote or virtual systems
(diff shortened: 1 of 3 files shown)


core-gl-kit/autogen.kit.d/base.yml
```diff
@@ -1299,6 +1299,74 @@ gitlab_freedesktop_dirlisting:
...
+          license: GPL-3
+          slot: '0'
+          homepage: https://gitlab.freedesktop.org/hadess/switcheroo-control/
+          gitlab_user: hadess
+          gitlab_repo: switcheroo-control
+          inherit:
+              - meson
+              - python-single-r1
+              - systemd
+          iuse: gtk-doc selinux systemd test
```
geaaru: I suggest to drop test USE and statically disable them in meson. I started to remove them from gnome too. The test process must be divided from ebuild process IMHO and keep things more clean in the ebuild.
`selinux` could be removed to probably.  The support of selinux is currently not available in Macaroni.

net-kit/autogen.kit.d/base.yml
```diff
@@ -2627,4 +2627,189 @@ gitlab_freedesktop_dirlisting:
...
+    github:
+      user: OpenPrinting
+      repo: libppd
+    assets:
+      - name: '{{ .Values.pn }}-{{ .Values.version }}.tar.xz'
+        url: 'https://github.com/OpenPrinting/{{ .Values.pn }}/releases/download/{{ .Values.version }}/{{ .Values.pn }}-{{ .Values.version }}.tar.xz'
+
+  packages:
+    - libppd:
+        transform:
```
geaaru: Maybe can we disable injection of the `rc|beta` releases? This could be done as exclude option and cleanup the trasforms.

text-kit/autogen.kit.d/base.yml
```diff
@@ -6,6 +6,10 @@ dev_github:
     category: app-text
     github:
       query: tags
+    transform:
+        - kind: string
+          match: "v"
```
geaaru: This is done automatically from `mark-devkit` IIRC. What is your error without it?

text-kit/autogen.kit.d/base.yml
```diff
@@ -25,6 +29,81 @@ dev_github:
...
+          - kind: string
+            match: 'release-qpdf-'
+            replace: ''
+        assets:
+          - name: '{{ .Values.pn }}-{{ .Values.version }}.tar.gz'
+            matcher: '{{ .Values.pn }}-{{ .Values.version }}.tar.gz'
+            url: 'https://github.com/qpdf/qpdf/releases/download/v{{ .Values.version }}/{{ .Values.pn }}-{{ .Values.version }}.tar.gz'
+          - name: '{{ .Values.pn }}-{{ .Values.version }}-doc.zip'
+            url: 'https://github.com/qpdf/qpdf/releases/download/v{{ .Values.version }}/{{ .Values.pn }}-{{ .Values.version }}-doc.zip'
+            final_name: '{{ .Values.pn }}-{{ .Values.version }}-doc.zip'
```
geaaru: This attribute doesn't exist. The name is directly take from the `name` attribute

text-kit/autogen.kit.d/base.yml
```diff
@@ -25,6 +29,81 @@ dev_github:
...
+            unpack: doc
```
geaaru: Typo? Do you means `use`?

text-kit/autogen.kit.d/base.yml
```diff
@@ -25,6 +29,81 @@ dev_github:
...
+        vars:
+          license: '|| ( Apache-2.0 Artistic-2 )'
+          slot: '0/$(ver_cut 1)'
+          desc: Command-line tool for structural, content-preserving transformation of PDF files
+          homepage: |
+            https://qpdf.sourceforge.io/
+            https://github.com/qpdf/qpdf/
+          inherit:
+            - cmake
+          iuse: doc examples gnutls test zopfli
```
geaaru: `test` could be removed. This simplify the dependency too.

text-kit/autogen.kit.d/base.yml
```diff
@@ -25,6 +29,81 @@ dev_github:
...
+          - name: '{{ .Values.pn }}-{{ .Values.version }}.tar.gz'
+            matcher: '{{ .Values.pn }}-{{ .Values.version }}.tar.gz'
+            url: 'https://github.com/qpdf/qpdf/releases/download/v{{ .Values.version }}/{{ .Values.pn }}-{{ .Values.version }}.tar.gz'
+          - name: '{{ .Values.pn }}-{{ .Values.version }}-doc.zip'
+            url: 'https://github.com/qpdf/qpdf/releases/download/v{{ .Values.version }}/{{ .Values.pn }}-{{ .Values.version }}-doc.zip'
+            final_name: '{{ .Values.pn }}-{{ .Values.version }}-doc.zip'
+            unpack: doc
+        vars:
+          license: '|| ( Apache-2.0 Artistic-2 )'
+          slot: '0/$(ver_cut 1)'
```
geaaru: Using helm template you can directly elaborate the first number of the version as macro. I added example in gnome IIRC.
Btw, IMHO, i don't think that it makes sense continue to follow Gentoo with subslotting. Probably you can just drop the slot attribute and use the default value with zero.

python-modules-kit/autogen.kit.d/python.yml
```diff
@@ -247,6 +247,20 @@ setuptools_modules:
...
+          license: MIT
+          desc: Library to parse gdb mi output and interact with gdb subprocesses
+          homepage: |
+            https://cs01.github.io/pygdbmi/
+            https://github.com/cs01/pygdbmi/
+            https://pypi.org/project/pygdbmi/
+          rdepend: '>=sys-devel/gdb-9.6'
+          body: |
+            EPYTEST_PLUGINS=()
+            distutils_enable_tests pytest
```
geaaru: Maybe can you avoid to set `body` and drop tests support?

net-kit/autogen.kit.d/base.yml
```diff
@@ -2627,4 +2627,189 @@ gitlab_freedesktop_dirlisting:
...
+      - name: '{{ .Values.pn }}-{{ .Values.version }}.tar.xz'
+        url: 'https://github.com/OpenPrinting/{{ .Values.pn }}/releases/download/{{ .Values.version }}/{{ .Values.pn }}-{{ .Values.version }}.tar.xz'
+
+  packages:
+    - libcupsfilters:
+        transform:
+          - kind: string
+            match: 'libcupsfilters '
+            replace: ''
+          - kind: string
```
geaaru: I think that you can exclude release rc*|beta|alpha directly with exclude and/or as regex in github section. It's better just use stable release.

kde-kit/merge.kit.d/base_autogen.yml
```diff
@@ -0,0 +1,31 @@
...
+        to: .github/FUNDING.yml
+
+  atoms_defaults:
+    max_versions: 3
+
+  atoms:
+    - pkg: app-crypt/qca
+    - pkg: dev-libs/plasma-wayland-protocols
+    - pkg: dev-libs/qcoro
+    - pkg: dev-libs/xerces-c
```
geaaru: Could be better add xerces-c in `dev-kit`? it seems a library not strictly used only for KDE.

PR #394, opened 2025-12-31
Autogens to enable kde6

linuxexplorer:
Bugs: macaroni-os/mark-issues#537
All new ebuilds have been tested locally and merge.


core-hw-kit/autogen.kit.d/base.yml
```diff
@@ -282,3 +282,114 @@ upower:
...
+          python_compat: python3+
+          desc: Makes power profiles handling available over D-Bus
+          homepage: https://gitlab.freedesktop.org/upower/power-profiles-daemon/
+          license: GPL-3+
+          inherit:
+            - bash-completion-r1
+            - meson
+            - python-single-r1
+            - systemd
+          iuse: bash-completion gtk-doc man selinux systemd test zsh-completion
```
geaaru: `test` could be removed together with the dependencies on BDEPEND and `python_check_deps` and `src_test`

core-hw-kit/autogen.kit.d/base.yml
```diff
@@ -282,3 +282,114 @@ upower:
...
+              meson_src_configure
+            }
+            src_test() {
+              # Dbus tests are prone to fail when run in parallel
+              MAKEOPTS="-j1" meson_src_test
+            }
+            src_install() {
+              meson_src_install
+              python_fix_shebang "${ED}"/usr/bin/powerprofilesctl
+              newinitd "${FILESDIR}/power-profiles-daemon.initd" power-profiles-daemon
```
geaaru, line 366: I'm not sure if it's really needed a initd daemon for this.
Alpine just patches the .dbus.service.in. See https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/community/power-profiles-daemon/0001-dbus-no-systemd.patch

We can try when the ebuild is created.

kde-kit/autogen.kit.d/base.yml
```diff
@@ -0,0 +1,488 @@
...
+                        $(qca_plugin_use gcrypt)
+                        $(qca_plugin_use gpg gnupg)
+                        $(qca_plugin_use logger)
+                        $(qca_plugin_use nss)
+                        $(qca_plugin_use pkcs11)
+                        $(qca_plugin_use sasl cyrus-sasl)
+                        $(qca_plugin_use softstore)
+                        $(qca_plugin_use ssl ossl)
+                        -DBUILD_TESTS=$(usex test)
+                )
```
geaaru, line 78: `test` to remove and probably indentation issues?

kde-kit/autogen.kit.d/base.yml
```diff
@@ -0,0 +1,488 @@
...
+    category: dev-libs
+    github:
+      user: danvratil
+      query: tags
+
+  packages:
+    - qcoro:
+        vars:
+            license: MIT
+            slot: '0'
```
geaaru: This is the default. You can remove it

kde-kit/autogen.kit.d/base.yml
```diff
@@ -0,0 +1,488 @@
...
+        >=kde-frameworks/extra-cmake-modules-5.115.0:*
+        virtual/pkgconfig
+
+  packages:
+    - phonon-vlc:
+        vars:
+          desc: VLC backend for the Phonon multimedia library
+          body: |
+            MY_PN="phonon-backend-vlc"
+            S="${WORKDIR}/${MY_PN}-${PV}"
```
geaaru: You can replace `MY_PN` and  `S`.
1) `S` could be defined as `s` vars eventually.
2) probably you can replace boh with this:
```yaml
vars:
   s: '${WORKDIR}/phonon-backend-vlc-{{ .Values.version }}'
```

kde-kit/autogen.kit.d/base.yml
linuxexplorer: Are you referring just to `slot: '0'`?

kde-kit/autogen.kit.d/base.yml
```diff
@@ -0,0 +1,488 @@
...
+                -DQT_MAJOR_VERSION=6
+                -DPHONON_BUILD_QT5=OFF
+                -DPHONON_BUILD_QT6=ON
+                -DKDE_INSTALL_USE_QT_SYS_PATHS=ON
+                -DKDE_INSTALL_DOCBUNDLEDIR="${EPREFIX}/usr/share/help"
+              )
+              cmake_src_configure
+            }
+
+pulseaudio_qt:
```
geaaru, line 364: With last releases of mark-devkit you can use Helm template over transform and matches. So probably you can use a single definition for both `pulseaudio_qt` and `polkit_qt` considering that using the same url.

So, something like this:
```yaml
# Example
mydef:
    generator: builtin-dirlisting
    defaults:
        dir:
           url: 'https://download.kde.org/stable/{{ .Values.pkgname | default .Values.pn }}/'
           matcher: '{{ .Values.pkgname | default .Values.pn }}-[0-9]+.tar.xz$'
         transform:
            - kind: string
              match: '{{ .Values.pkgname | default .Values.pn }}-'
              replace: ''
            - kind: string
              match: '.tar.xz'
              replace: ''
     packages:
        - pulseaudio-qt:
             ...
        - polkit-qt:
            ...
```

net-kit/autogen.kit.d/base.yml
```diff
@@ -2627,4 +2627,159 @@ gitlab_freedesktop_dirlisting:
...
+      - name: '{{ .Values.pn }}-{{ .Values.version }}.tar.xz'
+        url: 'https://github.com/OpenPrinting/{{ .Values.pn }}/releases/download/{{ .Values.version }}/{{ .Values.pn }}-{{ .Values.version }}.tar.xz'
+    vars:
+      license: Apache-2.0
+      slot: '0'
+      homepage: https://github.com/OpenPrinting/libcupsfilters
+      inherit:
+        - libtool
+      iuse: dbus exif jpeg +poppler +postscript png test tiff
+      restrict: '!test? ( test )'
```
geaaru: Also here `test` could be removed and set to `false` statically in the configure

net-kit/autogen.kit.d/base.yml
```diff
@@ -2627,4 +2627,159 @@ gitlab_freedesktop_dirlisting:
...
+  defaults:
+    files_dir: "{{ .Values.pn }}"
+    template: ../../templates/simple.tmpl
+    category: net-print
+    github:
+      user: OpenPrinting
+      repo: libcupsfilters
+    assets:
+      - name: '{{ .Values.pn }}-{{ .Values.version }}.tar.xz'
+        url: 'https://github.com/OpenPrinting/{{ .Values.pn }}/releases/download/{{ .Values.version }}/{{ .Values.pn }}-{{ .Values.version }}.tar.xz'
```
geaaru, line 2643: Cosmetic: When there assets and `builtin-github` you can also just use `matcher` of the tarball instead of the full path. See for example: https://github.com/macaroni-os/kit-fixups/blob/mark-unstable/core-server-kit/autogen.kit.d/telemetry.yml#L17.

It's fine this way too. Just sharing an alternative way for github packages.

kde-kit/autogen.kit.d/qca/qca-disable-pgp-test.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/unittest/CMakeLists.txt b/unittest/CMakeLists.txt
+index f4d9eb9..67ef281 100644
+--- a/unittest/CMakeLists.txt
++++ b/unittest/CMakeLists.txt
+@@ -23,7 +23,7 @@ add_subdirectory(keylengthunittest)
+ add_subdirectory(keystore)
+ add_subdirectory(macunittest)
+ add_subdirectory(metatype)
+-add_subdirectory(pgpunittest)
++# add_subdirectory(pgpunittest)
```
geaaru: I think that this Gentoo-style patch could be replaced with a simple sed in `src_prepare`.
Something, like this:
```
sed -e '/^add_subdirectory\(pgpunittest/d' unittest/CMakeLists.txt -i || die
```

kde-kit/autogen.kit.d/base.yml
geaaru: Yeah, it's the default. You can drop the line. ;)

dev-kit/autogen.kit.d/base.yml
```diff
@@ -1850,3 +1850,139 @@ appstream:
...
+        url: 'https://archive.apache.org/dist/xerces/c/3/sources/{{ .Values.pn }}-{{ .Values.version }}.tar.xz'
+    vars:
+      license: Apache-2.0
+      slot: '0/$(ver_cut 1-2)'
+      homepage: https://xerces.apache.org/xerces-c/
+      inherit:
+        - cmake
+        - flag-o-matic
+        - prefix
+      iuse: cpu_flags_x86_sse2 curl doc examples iconv icu static-libs test threads
```
geaaru: `test` could be removed

geaaru:
Moving ahead this PR that fixes also:
* https://github.com/macaroni-os/mark-issues/issues/138
* https://github.com/macaroni-os/mark-issues/issues/153

geaaru:
Thank you very much

PR #395, opened 2026-01-12
Initial_kde6_autogens_set_profile

linuxexplorer:
Initial commit for kde6 autogens, sets, and profile.


core-kit/sets/kde-plasma
```diff
@@ -0,0 +1,54 @@
...
+net-misc/networkmanager[elogind]
+
+sys-apps/dbus
+
+sys-auth/elogind[pam]
+sys-auth/polkit
+
+sys-fs/udisks:2
+
+x11-misc/sddm
```
geaaru, line 54: I dunno if it makes sense add `sddm` inside the KDE sets. I can start KDE session having GDM too. Probably we could avoid this?

kde-kit/autogen.kit.d/base.yml
```diff
@@ -205,6 +205,79 @@ stable_kde_stuff:
...
+        category: dev-libs
+        vars:
+          homepage: https://invent.kde.org/graphics/kdiagram
+          desc: Powerful libraries (KChart, KGantt) for creating business diagrams
+          license: GPL-2
+          slot: 6
+          inherit:
+            - kde6
+          iuse: examples test
+          required_use: test? ( examples )
```
geaaru, line 218: Please, drop `test` IUSE

kde-kit/autogen.kit.d/kde-apps.yaml
```diff
@@ -0,0 +1,6032 @@
...
+       - name: "{{ .Values.pn }}-{{ .Values.version}}.tar.xz"
+         url: https://download.kde.org/stable/release-service/{{ .Values.release }}/src/{{ .Values.pn }}-{{ .Values.version}}.tar.xz
+    dir:
+        url: "https://download.kde.org/stable/release-service/"
+        matcher: "^25.*"
+
+  packages:
+    - akonadi:
+        vars:
+          inherit: kde6 xdg
```
geaaru, line 26: This is wrong. `inherit` is an array.
```yaml
inherit:
- kde6
- xdg
```

kde-kit/autogen.kit.d/kde-apps.yaml
```diff
@@ -0,0 +1,6032 @@
...
+          iuse: "tools +webengine xml"
+          desc: "Storage service for PIM data and libraries for PIM apps"
+          homepage: https://community.kde.org/KDE_PIM/akonadi
+          common_depend: |
```
geaaru: This is `cdepend` if you use the `simple.tmpl` and later `${CDEPEND}}`

kde-kit/autogen.kit.d/kde-apps.yaml
```diff
@@ -0,0 +1,6032 @@
...
+                kde-apps/kaccounts-integration:{{ .Values.slot }}
+                >=net-libs/accounts-qt-1.17[qt6(+)]
+            )
+            xml? ( dev-libs/libxml2:= )
+          depend: |
+            ${COMMON_DEPEND}
+            dev-libs/libxslt
+          rdepend: |
+            ${COMMON_DEPEND}
+            !<app-office/merkuro-25.07.80
```
geaaru: We haven't a `merkuro` package IIRC. This could be removed.

kde-kit/autogen.kit.d/base.yml
```diff
@@ -205,6 +205,79 @@ stable_kde_stuff:
...
+            kde-frameworks/kiconthemes:6
+            kde-frameworks/kio:6
+            kde-frameworks/kitemviews:6
+            kde-frameworks/kservice:6
+            kde-frameworks/ktextwidgets:6
+            kde-frameworks/kwidgetsaddons:6
+            kde-frameworks/sonnet:6
+            kde-frameworks/syntax-highlighting:6
+            speech? ( dev-qt/qtspeech:6 )
+          rdepend: |
```
geaaru, line 268: RDEPEND is defined before DEPEND in the template. This means that you need define package in RDEPEND and later using `${RDEPEND}` in DEPEND else Portage ebuild script doesn't work correctly.

kde-kit/autogen.kit.d/base.yml
```diff
@@ -205,6 +205,79 @@ stable_kde_stuff:
...
+            url: 'https://download.kde.org/stable/ktextaddons/{{ .Values.pn }}-{{ .Values.version }}.tar.xz'
+        vars:
+          homepage: https://invent.kde.org/libraries/ktextaddons
+          desc: Various text handling addons
+          license: GPL-2
+          slot: 6
+          inherit:
+            - kde6
+          iuse: speech
+          restrict: test
```
geaaru, line 249: not needed. We haven't test. You can remove it.

kde-kit/autogen.kit.d/base.yml
```diff
@@ -205,6 +205,79 @@ stable_kde_stuff:
...
+            kde-frameworks/kitemviews:6
+            kde-frameworks/kservice:6
+            kde-frameworks/ktextwidgets:6
+            kde-frameworks/kwidgetsaddons:6
+            kde-frameworks/sonnet:6
+            kde-frameworks/syntax-highlighting:6
+            speech? ( dev-qt/qtspeech:6 )
+          rdepend: |
+            ${DEPEND}
+            !${CATEGORY}/${PN}:5
```
geaaru, line 270: I suggest using helm engine here instead of Portage variables:
```yaml
!{{ .Values.category }}/{{ .Values.pn }}:5
```

kde-kit/autogen.kit.d/kde-apps.yaml
```diff
@@ -0,0 +1,6032 @@
...
+            kde-frameworks/kwidgetsaddons:{{ .Values.slot }}
+            kde-frameworks/kwindowsystem:{{ .Values.slot }}
+            kde-frameworks/kxmlgui:{{ .Values.slot }}
+            webengine? (
+                kde-apps/kaccounts-integration:{{ .Values.slot }}
+                >=net-libs/accounts-qt-1.17[qt6(+)]
+            )
+            xml? ( dev-libs/libxml2:= )
+          depend: |
+            ${COMMON_DEPEND}
```
geaaru: Same here: `${CDEPEND}`

kde-kit/autogen.kit.d/kde-apps.yaml
```diff
@@ -0,0 +1,6032 @@
...
+          rdepend: |
+            ${DEPEND}
+          body: |
+            src_prepare() {
+                kde6_src_prepare
+            }
+
+    - akonadi-contacts:
+        vars:
+          inherit: kde6
```
geaaru: `inherit` to convert in array

kde-kit/autogen.kit.d/kde-apps.yaml
```diff
@@ -0,0 +1,6032 @@
...
+          rdepend: |
+            ${DEPEND}
+          body: |
+            src_prepare() {
+                kde6_src_prepare
+            }
+
+    - akonadi-import-wizard:
+        vars:
+          inherit: kde6
```
geaaru: `inherit` to convert in array. Maybe you can add this `inherit` as default and drop the field where is equal.

kde-kit/autogen.kit.d/kde-apps.yaml
```diff
@@ -0,0 +1,6032 @@
...
+            kde-frameworks/karchive:{{ .Values.slot }}
+            kde-frameworks/kconfig:{{ .Values.slot }}
+            kde-frameworks/kcontacts:{{ .Values.slot }}
+            kde-frameworks/kcoreaddons:{{ .Values.slot }}
+            kde-frameworks/kcrash:{{ .Values.slot }}
+            kde-frameworks/kdbusaddons:{{ .Values.slot }}
+            kde-frameworks/ki18n:{{ .Values.slot }}
+            kde-frameworks/kio:{{ .Values.slot }}
+            kde-frameworks/kwidgetsaddons:{{ .Values.slot }}
+          rdepend: |
```
geaaru: Here same issue. RDEPEND is before DEPEND and you need to invert the logic.

kde-kit/autogen.kit.d/kde-apps.yaml
```diff
@@ -0,0 +1,6032 @@
...
+            ${DEPEND}
+          body: |
+            src_prepare() {
+                kde6_src_prepare
+            }
+
+    - akonadi-search:
+        vars:
+          inherit: kde6
```
geaaru: `inherit` to drop. Set this var as default in the `defaults` section.

kde-kit/autogen.kit.d/kde-apps.yaml
```diff
@@ -0,0 +1,6032 @@
...
+            kde-frameworks/kconfig:{{ .Values.slot }}
+            kde-frameworks/kcoreaddons:{{ .Values.slot }}
+            kde-frameworks/kcrash:{{ .Values.slot }}
+            kde-frameworks/kdbusaddons:{{ .Values.slot }}
+            kde-frameworks/ki18n:{{ .Values.slot }}
+            kde-frameworks/kxmlgui:{{ .Values.slot }}
+          rdepend: |
+            ${DEPEND}
+          body: |
+            src_prepare() {
```
geaaru: Maybe you can set this body as default and define it only in the packages with differences.

kde-kit/autogen.kit.d/kde-apps.yaml
```diff
@@ -0,0 +1,6032 @@
...
+                kde6_src_prepare
+            }
+
+    - kmplot:
+        vars:
+          inherit: kde6 xdg
+          desc: "Mathematical function plotter"
+          homepage: https://apps.kde.org/kmplot/
```
geaaru, line 3134: Maybe you can set this in the default section:
```
homepage: 'https://apps.kde.org/{{ .Values.pn }}/'
```
and drop this field in every package that has this value

kde-kit/merge.kit.d/kde-plasma.yaml
```diff
@@ -0,0 +1,89 @@
...
+    - pkg: kde-plasma/flatpak-kcm
+    - pkg: kde-plasma/kactivitymanagerd
+    - pkg: kde-plasma/kde-cli-tools
+    - pkg: kde-plasma/kde-cli-tools-common
+    - pkg: kde-plasma/kde-gtk-config
+    - pkg: kde-plasma/kdecoration
+    - pkg: kde-plasma/kdeplasma-addons
+    - pkg: kde-plasma/kgamma
+    - pkg: kde-plasma/kglobalacceld
+    - pkg: kde-plasma/kinfocenter-6.
```
geaaru: typo at the end?

kde-kit/merge.kit.d/kde-plasma.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+      - file: ../../.github/FUNDING.yml
+        to: .github/FUNDING.yml
+
+  atoms_defaults:
+    max_versions: 2
+
+  atoms:
+
+    - pkg: kde-plasma/aurorae
```
geaaru: maybe it's better to add `:6` related to the slot

geaaru (changes requested):

core-kit/sets/kde-plasma
linuxexplorer: We leave it to the user to choose a display manager of start from terminal? I would prefer sddm being kept in the set. From plasma-6.6, there will be a kde display manager, but it has a hard dependency on systemd :(
Is it possible to have a 'one-of-several' option (sddm | gdm) or suggestion?

kde-kit/autogen.kit.d/base.yml
linuxexplorer: The package will not merge correctly without this arrangement - the alternative is to have a patch or sed

kde-kit/autogen.kit.d/kde-apps.yaml
linuxexplorer: Oh, seemed to work:
```
# Distributed under the terms of the GNU General Public License v2
# Autogen by MARK Devkit

EAPI=7
inherit kde6 xdg
```

core-kit/sets/kde-plasma
geaaru: yeah, Systemd begin to be a pain. On Phoenix is not an issue because profiles are not used but having a fixed XDM by default always installed if someone wants other is not so good for me. btw, we can start in this way. I need time to investigate in new XDM

kde-kit/autogen.kit.d/base.yml
geaaru: This because `kde6.eclass` want it probably. Can you set always tests to off/disable there?

core-kit/sets/kde-plasma
cuantar: would be good if XDM satisfies the requirement.  I don't know how to do that with a set, though.

kde-kit/autogen.kit.d/kde-apps.yaml
linuxexplorer: I'll change it to depend

kde-kit/autogen.kit.d/base.yml
linuxexplorer: I don't believe so, no. I tried a few different ways to disable tests but without success.
Speaking of eclass, what is the best way to add them to mark-unstable? Is it possible to have `core-kit/eclass` with just these 2 eclass or would I need to copy all existing eclass into there too?

kde-kit/autogen.kit.d/kde-apps.yaml
geaaru: eheh, probably because it see it like a single string and iterate on it. Honestly, if it's possible it's better keep a single format, will be more easy to display on an interface web if we convert them in json for a website somewhere.

geaaru: But good to know that it works in any case

kde-kit/autogen.kit.d/base.yml
geaaru: IMHO, it's better keep all eclasses in a single place on `core-kit/eclass` ;)

geaaru: ah sorry, there is already a single place under the root path of kit-fixups. See https://github.com/macaroni-os/kit-fixups/tree/mark-unstable/eclass. This directory is already synced in core-kit from mark-devkit bot

core-kit/sets/kde-plasma
geaaru: If I prefer the `sets` for set a common list of packages to have a KDE base system, GNOME base system, etc. doesn't means that we can't have meta ebuild anymore. What I prefer is try to reduce the over-use of the ebuild/eclass that is present in Gentoo. I'm more oriented in a *divide and conquer* way where also other tools can help on install stuff like *mix-in*, *flavor* do already in part to set automatically specific USEs (for example with `whip`, `macaronictl`).
So, if an XDM is mandatory for the setup we can also create a `virtual/xdm` packages based on USEs that are later configured with different values based on mix-in etc. At least for now until we will replace Portage/ego stuff ;)

eclass/kde6.eclass
```diff
@@ -0,0 +1,714 @@
...
+		if [[ "${pair}" = "${CATEGORY}-${PV}" ]]; then
+			return 0
+		fi
+	done
+
+	return 1
+}
+
+# Determine fetch location for released tarballs
+_calculate_src_uri() {
```
geaaru: The `_calculate_src_uri` could be removed. We generate the URLs with mark-devkit. ;)

eclass/kde6.eclass
```diff
@@ -0,0 +1,714 @@
...
+		esac
+	fi
+
+	if _kde_is_unreleased ; then
+		RESTRICT+=" fetch"
+	fi
+}
+
+# Determine fetch location for live sources
+_calculate_live_repo() {
```
geaaru: Same for `_calculate_live_repo`. We don't generate 9999 packages

eclass/kde6.eclass
```diff
@@ -0,0 +1,714 @@
...
+	debug-print-function ${FUNCNAME} "$@"
+
+	cmake-utils_src_compile "$@"
+}
+
+# @FUNCTION: kde6_src_test
+# @DESCRIPTION:
+# Wrapper for cmake-utils_src_test with extra logic for magic handling of dbus
+# and virtualx.
+kde6_src_test() {
```
geaaru: Probably we can drop `kde6_src_test`.

kde-kit/autogen.kit.d/kde-frameworks.yaml
```diff
@@ -426,15 +423,15 @@ kde-frameworks:
           inherit: kde6 xdg
           desc: "Framework for solving common problems such as caching, randomisation, and more"
           iuse: "dbus"
+          rdepend: |
+            dev-qt/qttranslations:{{ .Values.slot }}
```
geaaru, line 427: Probably here there is an error. In the previous implementation with:
```
rdepend: |
   ${DEPEND}
   dev-qt/qttranslations:{{ .Values.slot }}
```
RDEPEND was with qtbase, qtdeclerative, qttools and virtual/libudev.
Now RDEPEND contains only qttranslations. Is this correct?
I think that is correct instead that sys-kernel/linux-headers is only a DEPEND or BDEPEND dependency.

kde-kit/autogen.kit.d/kde-plasma.yaml
```diff
@@ -186,7 +185,10 @@ kde-plasma:
           inherit: kde6 xdg
           iuse: "+firmware flatpak snap telemetry webengine"
           desc: "KDE Plasma resources management GUI"
+          rdepend: |
+            snap? ( app-containers/snapd )
```
geaaru: Does we really want `snap`? :scream:

kde-kit/autogen.kit.d/kde-plasma.yaml
```diff
@@ -1181,7 +1180,17 @@ kde-plasma:
...
           desc: "Flexible, composited Window Manager for windowing systems on Linux"
+          rdepend: |
+            !kde-plasma/kdeplasma-addons:5
+            !<kde-plasma/kwin-6.3.80
+            kde-frameworks/kirigami:6
+            kde-frameworks/kitemmodels:6
+            kde-plasma/aurorae:6
+            kde-plasma/libplasma:6
+            sys-apps/hwdata
+            >=x11-base/xwayland-23.1.0
```
geaaru, line 1189: Is this correct? Is no more supported x11 on KDE6? If yes, probably we can add a `virtual/gui-server` that will be something like this:
```
IUSE="wayland x11 xlibre"
RDEPEND="
x11? ( x11-base/xorg-server )
wayland? ( x11-base/xwayland )
xlibre? ( x11-base/xlibre-server )
"
WDYT?
```

kde-kit/autogen.kit.d/kde-plasma.yaml
linuxexplorer: No, I'll remove it :+1:

kde-kit/autogen.kit.d/kde-plasma.yaml
linuxexplorer: x11 is still support on kde6 for this year, at least: https://blogs.kde.org/2025/11/26/going-all-in-on-a-wayland-future/
xorg-server is a dependency in plasm-desktop and sddm ebuilds but I will add it to kwin-x11 too :+1:

kde-kit/autogen.kit.d/kinfocenter/kcm-about-distrorc
```diff
@@ -0,0 +1,4 @@
+[General]
+Name=Funtoo Linux
```
geaaru: Maybe, `Macaroni OS Linux` :)

kde-kit/autogen.kit.d/plasma-workspace/10-agent-shutdown.sh
geaaru, line 1: This file seems useless. All code is commented. Is this a legacy file?

geaaru: Ah, maybe is this a stub about how users could add post-shutdown hooks?

kde-kit/autogen.kit.d/kinfocenter/kcm-about-distrorc
linuxexplorer: This has already been fixed ;-) https://github.com/macaroni-os/kit-fixups/pull/395/changes/00434c448d8e1fc54a8fc9ea80473238305ff9d3

kde-kit/autogen.kit.d/plasma-workspace/10-agent-shutdown.sh
linuxexplorer: Yes, it could be included with everything commented out or with the relevant parts uncommented

kde-kit/autogen.kit.d/kinfocenter/kcm-about-distrorc
geaaru: Sorry, this PR begin to be too big and difficult to read with all these commits :)

If you agree i think that we can move ahead and eventually add fixes after a good testing in mark-unstable. WDYT?

linuxexplorer: That sounds good to me :-)

geaaru:
Going ahead then! Thank you very much for all this work! :champagne:

PR #396, opened 2026-01-14
[core-kit, core-server-kit, dev-kit] Autogen `libcap`, `sssd`, and `ding-libs`

cuantar:
Packages are in different kits:
- core-kit: libcap
- core-server-kit: sssd
- dev-kit: ding-libs

The goal here is `sssd`; the others are dependencies.

### DIFF
--- a/core-kit/autogen.kit.d/core.yml
+++ b/core-kit/autogen.kit.d/core.yml
@@ -855,3 +855,79 @@ libxml2:
     - libxml2:
 
 
+libcap:
+  generator: builtin-dirlisting
+  generator_opts:
+    rate_limiter: "1"
+  defaults:
+    files_dir: "{{ .Values.pn }}"
+    category: sys-libs
+    template: ../../templates/simple.tmpl
+    dir:
+      url: https://www.kernel.org/pub/linux/libs/security/linux-privs/libcap2/
+      matcher: "{{ .Values.pn }}-2.*.tar.gz$"
+
+    transform:
+      - kind: string
+        match: "{{ .Values.pn }}-"
+        replace: ""
+      - kind: string
+        match: ".tar.gz"
+        replace: ""
+    assets:
+      - name: "{{ .Values.pn }}-{{ .Values.version }}.tar.gz"
+        url: https://git.kernel.org/pub/scm/libs/libcap/libcap.git/snapshot/{{ .Values.pn }}-{{ .Values.version }}.tar.gz
+
+    vars:
+      desc: "POSIX 1003.1e capabilities"
+      homepage: "https://sites.google.com/site/fullycapable/"
+      license: "|| ( GPL-2 BSD ) pam? ( || ( LGPL-2+ BSD ) )"
+      slot: '0'
+      inherit:
+        - pam
+        - toolchain-funcs
+      iuse: pam +static-libs tools
+      rdepend: |
+        pam? ( sys-libs/pam )
+      depend: |
+        ${RDEPEND}
+        sys-kernel/linux-headers
+      bdepend:
+        sys-apps/diffutils
+        tools? ( dev-lang/go )
+      body: |
+        run_emake() {
+          local args=(
+          	exec_prefix="${EPREFIX}"
+          	lib_prefix="${EPREFIX}/usr"
+          	lib="$(get_libdir)"
+          	prefix="${EPREFIX}/usr"
+          	PAM_CAP="$(usex pam yes no)"
+          	DYNAMIC=yes
+          	GOLANG="$(usex tools yes no)"
+          )
+          emake "${args[@]}" "$@"
+        }
+        src_compile() {
+          local BUILD_CC
+          tc-export_build_env BUILD_CC
+          run_emake
+        }
+        src_install() {
+          run_emake DESTDIR="${D}" install
+          gen_usr_ldscript -a cap
+          if ! use static-libs; then
+            rm "${ED}"/usr/$(get_libdir)/lib{cap,psx}.a || die
+          fi
+          # install pam plugins ourselves
+          if [[ -d "${ED%/}"/usr/$(get_libdir)/security ]] ; then
+            rm -rf "${ED}"/usr/$(get_libdir)/security || die
+          fi
+          if use pam; then
+            dopammod pam_cap/pam_cap.so
+            dopamsecurity '' pam_cap/capability.conf
+          fi
+        }
+
+  packages:
+    - libcap:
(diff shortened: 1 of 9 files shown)


cuantar:
Marked as draft until PR for adding the `sssd` user/group to entities in `macaroni-commons` repo is ready.

geaaru:
It sounds all good. Thanks. Just a note.... probably we can stop to use EPREFIX that was an old feature of Portage in order to have an installation over a custom path. Now with `chroot`, containers, etc. honestly, i dunno if it makes sense yet.

cuantar:
Changed the url to https://www.kernel.org/pub/linux/libs/security/linux-privs/libcap2/ as discussed in Discord.

core-server-kit/autogen.kit.d/base.yml
```diff
@@ -320,3 +321,279 @@ eudev:
...
+        selinux? (
+          sys-libs/libselinux
+          sys-libs/libsemanage
+        )
+        systemd? (
+          sys-apps/systemd:=
+          sys-apps/util-linux
+        )
+        systemtap? ( dev-debug/systemtap )
+      rdepend: |
```
geaaru: Same error reported for KDE 6 :)
The template have RDEPEND before DEPEND as position. This means that when Portage load the file the DEPEND will be empty and not with the content defined on `depend` attribute.
The position in the yaml is not directly connected to the position in the template.
So, in summary it's needed to use CDEPEND or add all dependencies in RDEPEND and later use:
```yaml
depend: '${RDEPEND}'
```

core-server-kit/autogen.kit.d/base.yml
```diff
@@ -320,3 +321,279 @@ eudev:
...
+        ${DEPEND}
+        passkey? ( sys-apps/pcsc-lite[policykit] )
+        selinux? ( sec-policy/selinux-sssd )
```
geaaru: Atm we haven't SELINUX support and/or packages. I think that we can remove this line and dependencies for `selinux` USE

dev-kit/autogen.kit.d/base.yml
```diff
@@ -1981,3 +1981,33 @@ xerces_c:
...
+      - kind: string
+        match: '{{ .Values.pn }}-'
+        replace: ''
+    assets:
+      - name: '{{ .Values.pn }}-{{ .Values.version }}.tar.gz'
+    vars:
+      desc: \"DING is not GNU\" helper libraries for SSSD and FreeIPA
+      rdepend:
+      depend: |
+        ${RDEPEND}
```
geaaru: Typo? `rdepend` is empty and later used in `depend`.
Probably you can convert these lines in:
```yaml
bdepend: |
    virtual/pkgconfig
 ```
 and drop both `rdepend` and `depend` if there aren't rows missed.

File: core-kit/autogen.kit.d/core.yml [hunk 4]
```diff
@@ -855,3 +855,79 @@ libxml2:
...
+      license: "|| ( GPL-2 BSD ) pam? ( || ( LGPL-2+ BSD ) )"
+      slot: '0'
+      inherit:
+        - pam
+        - toolchain-funcs
+      iuse: pam +static-libs tools
+      pdepend: |
+        pam? ( sys-libs/pam )
+      depend: |
+        ${PDEPEND}
```
geaaru: This contains two issues:
1. PDEPEND is after DEPEND in template so will never be evaluate correctly.
2. Probably this is a typo because in the the ebuild in the tree `pam? ( virtual/pam )` is in RDEPEND. Is`RDEPEND` and not PDEPEND here?

core-server-kit/autogen.kit.d/base.yml
cuantar: done!

core-server-kit/autogen.kit.d/base.yml
cuantar: done, removed the `selinux` flag and added `--without-selinux` to config options

dev-kit/autogen.kit.d/base.yml
cuantar: yes, that seems right; I wasn't sure about whether it would be okay to leave them undefined.  removed/changed to `bdepend`

core-kit/autogen.kit.d/core.yml
cuantar: I think you are correct.  Fixed.

cuantar:
I think this is good, except for that I need to set the `uid`:`gid` of the new user and group that are created, to match in `entities`.

core-server-kit/autogen.kit.d/base.yml
```diff
@@ -320,3 +321,275 @@ eudev:
...
+        - 'PLOCALES="ca de es fr ja ko pt_BR ru sv tr"'
+        - 'PLOCALES_BIN="${PLOCALES} bg cs eu fi hu id it ka nb nl pl pt tg zh_TW zh_CN"'
+        - 'PLOCALE_BACKUP="sv"'
+        - 'CONFIG_CHECK="~KEYS"'
+
+      body: |
+        pkg_setup() {
+          linux-info_pkg_setup
+          python-single-r1_pkg_setup
+          enewgroup sssd
```
geaaru: If you want keep aligned the user and group with entities-catalog you need to add the [gid] after the name and for the user replace -1 with the [uid].
It seems so reading the `user.eclass`.

cuantar:
Done!  Should be okay to merge now.  Has been tested and seems to do what it needs to do, without breaking anything.

cuantar:
(Added explicit version requirement to `libcap` dep, so that the new version will be pulled in before trying to merge `sssd`)

core-kit/merge.kit.d/core.yml
```diff
@@ -371,7 +371,6 @@ target:
     - pkg: sys-devel/make
     - pkg: sys-devel/patch
 
-    - pkg: sys-libs/libcap
```
geaaru, line 372: This line needs to be added in `merge.kit.d/core_autogen.yml` else the mark-devkit bot skip the package.

geaaru:
You need to add `sssd` and `ding-libs` packages in the specific merge script file.

cuantar:
Okay, done.  Needed to do `libcap` as well, since it was in our tree but wasn't autogen.

core-kit/merge.kit.d/core.yml
cuantar: Ah, I guess I intended to move it but only did it halfway.  Caught it myself too.  Thanks for being careful!

geaaru:
Thanks! :rocket:

PR #397, opened 2026-01-21
Add_meta_flags_to_kde6_profile

linuxexplorer:
Add the + use flags from plasma-meta to kde6 profile


core-kit/profiles/funtoo/1.0/linux-gnu/mix-ins/kde-plasma-6/package.use
```diff
@@ -168,3 +168,14 @@ kde-frameworks/sonnet qml
 kde-frameworks/kwallet runtime
 app-office/libreoffice-bin -kde
 app-text/poppler nss
+
+# meta flags
+*/* browser-integration
```
geaaru, line 173: I never see this syntax in `package.use`. Is not equivalent to add these USE in `make.defaults` ?

geaaru:
Add reference to an issue related to this PR: https://github.com/macaroni-os/mark-issues/issues/603

PR #405, opened 2026-03-09
[lang-kit] Optimize deno build settings and stop forcing V8 source builds

org-tekeli-borisp:
## Summary

Adjust `net-libs/deno` build settings to use explicit Cargo/Ninja parallelism controls and stop forcing V8 source builds.

## What changed

- add `src_configure()` to set:
  - `MAKEOPTS="-j4"`
  - `cargo_gen_config`
- replace `cargo_src_compile` with a direct:
  - `cargo build --release`
- stop forcing:
  - `V8_FROM_SOURCE=1`
- add explicit build tuning in `src_compile()`:
  - `CARGO_BUILD_JOBS=4`
  - `CARGO_INCREMENTAL=0`
  - `CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1`
  - `NINJAFLAGS="-j4"`

## Why

Forcing `V8_FROM_SOURCE=1` sent the build down the Chromium/V8 GN/Ninja source-build path, which brought in extra complexity and failures unrelated to building Deno itself.

Using the normal `rusty_v8` path is simpler and avoids those source-build-specific issues.

This change also makes resource usage more predictable by setting Cargo/Ninja parallelism explicitly instead of relying on previous eclass-driven behavior.

## Rationale

- `CARGO_INCREMENTAL=0` reduces build overhead for one-shot package builds
- `CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1` trades some build speed for lower memory pressure and more stable release builds
- explicit `MAKEOPTS`, `CARGO_BUILD_JOBS`, and `NINJAFLAGS` make the package behavior easier to reason about

## Testing

Tested by building:

- `net-libs/deno-2.7.4`

with the updated autogen recipe.

## Notes

This PR intentionally removes the forced V8 source-build behavior rather than trying to maintain extra workarounds for the GN/Ninja path.

### DIFF
--- a/lang-kit/autogen.kit.d/base.yml
+++ b/lang-kit/autogen.kit.d/base.yml
@@ -677,10 +677,17 @@ deno:
           rm -rf "${S}" || true
           mv {{ .Values.github_user }}-{{ .Values.github_repo }}-* "${S}"
         }
+        src_configure() {
+          export MAKEOPTS="-j1"
+          cargo_gen_config
+        }
         src_compile() {
-          # Don't try to fetch prebuilt V8, build it instead
-          export V8_FROM_SOURCE=1
-          cargo_src_compile
+          export CARGO_BUILD_JOBS=1
+          export CARGO_INCREMENTAL=0
+          export CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1
+          export NINJAFLAGS="-j1"
+        
+          cargo build --release || die "cargo build failed"
         }
         src_install() {
           # Install the binary directly, cargo install doesn't work on workspaces
@@ -689,4 +696,3 @@ deno:
 
   packages:
     - deno:
-


lang-kit/autogen.kit.d/base.yml
```diff
@@ -677,10 +677,17 @@ deno:
           rm -rf "${S}" || true
           mv {{ .Values.github_user }}-{{ .Values.github_repo }}-* "${S}"
         }
+        src_configure() {
+          export MAKEOPTS="-j4"
```
geaaru: Maybe could this be handled by `memsaver_src_configure` function?

org-tekeli-borisp: changed to conservative 1

PR #406, opened 2026-03-09
[python-build-tools-kit] dev-util/maturin Increase `MEMSAVER_FACTOR` from 2G to 10G to optimize build resource allocation

org-tekeli-borisp:


geaaru:
I'm begin to think that we need to review the memsaver eclass with a better logic :)

PR #408, opened 2026-03-13
[core-kit] Autogen `sys-libs/glibc` (first pass)

cuantar:
Also add autogens for
- [core-kit] `sys-apps/locale-gen`
- [dev-kit] `sys-apps/man-pages`
- [security-kit] `app-misc/pax-utils`

### DIFF
--- a/core-kit/autogen.kit.d/templates/glibc.tmpl
+++ b/core-kit/autogen.kit.d/templates/glibc.tmpl
@@ -1,0 +1,1458 @@
+# Distributed under the terms of the GNU General Public License v2
+# Autogen by MARK Devkit
+
+EAPI={{ .Values.eapi | default "7" }}
+
+{{- if .Values.python_compat | default false }}
+PYTHON_COMPAT=( {{ .Values.python_compat }})
+{{- end }}
+{{- if .Values.extra_envs | default false }}
+{{- range $p := .Values.extra_envs }}
+{{ $p }}
+{{- end }}
+{{- end }}
+
+{{- if .Values.inherit | default false }}
+inherit {{ .Values.inherit | join " " }}
+{{- end }}
+
+DESCRIPTION="{{ .Values.desc }}"
+{{- if .Values.homepage | default false }}
+HOMEPAGE="{{ .Values.homepage }}"
+{{- end }}
+{{- $assets_len := .Values.artefacts | len }}
+{{- if gt $assets_len 1 }}
+SRC_URI="{{- range $k, $v := .Values.artefacts }}
+{{- if eq $v.Use "" }}
+{{ $v.SrcUri | first }} -> {{ $v.Name }}
+{{- else }}
+{{ $v.Use }}? ( {{ $v.SrcUri | first }} -> {{ $v.Name }} )
+{{- end }}
+{{- end }}"
+{{- else }}
+{{- if .Values.src_uri | default false }}
+SRC_URI="{{ .Values.src_uri }}"
+{{- end }}
+{{- end }}
+
+{{- if .Values.license | default false }}
+LICENSE="{{ .Values.license }}"
+{{- end }}
+SLOT="{{ .Values.slot | default 0 }}"
+KEYWORDS="{{ .Values.keywords | default "*" }}"
+
+{{- if .Values.patches | default false }}
+PATCHES=(
+	{{- range $p := .Values.patches }}
+	"${FILESDIR}/{{ $p }}"
+	{{- end }}
+)
+{{- end }}
+
+{{- if .Values.iuse | default false }}
+IUSE="{{ regexReplaceAll "\\n" .Values.iuse "\n\t" }}"
+{{- end }}
+{{- if .Values.required_use | default false }}
+REQUIRED_USE="{{ .Values.required_use }}"
+{{- end }}
+{{- if .Values.restrict | default false }}
+RESTRICT="{{ .Values.restrict }}"
+{{- end }}
+
+{{- if .Values.depend | default false }}
+DEPEND="{{ regexReplaceAll "\\n" .Values.depend "\n\t" }}
+"
+{{- end }}
+{{- if .Values.bdepend | default false }}
+BDEPEND="{{ regexReplaceAll "\\n" .Values.bdepend "\n\t" }}
+"
+{{- end }}
+{{- if .Values.rdepend | default false }}
+RDEPEND="{{- regexReplaceAll "\\n" .Values.rdepend "\n\t" }}
+"
+{{- end }}
+{{- if .Values.pdepend | default false }}
+PDEPEND="{{ regexReplaceAll "\\n" .Values.pdepend "\n\t" }}
+"
+{{- end }}
+{{- if .Values.s | default false }}
+S="{{ .Values.s }}"
+{{- end }}
+
+# Bumping notes: https://wiki.gentoo.org/wiki/Project:Toolchain/sys-libs/glibc
+# Please read & adapt the page as necessary if obsolete.
+
+TMPFILES_OPTIONAL=1
+
+# Gentoo patchset
+PATCH_VER={{ .Values.patch_version }}
+PATCH_DEV={{ .Values.patch_dev }}
+
+# gcc mulitilib bootstrap files version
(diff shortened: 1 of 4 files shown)


cuantar:
Will mark as ready once local tests succeed.

cuantar:
The `locale-gen` executable that used to be part of `glibc` now is its own package, like `libxcrypt`; so we need to autogen `sys-apps/locale-gen` also.

... but, the Gentoo ebuild for `locale-gen` requires `perl-5.36` and `dev-perl/File-Slurper`, both not in our tree. :/

Need to investigate whether that code is the same as for the `locale-gen` in other distros.  In Debian, it's in the `locales` package.

cuantar:
Short answer: `locale-gen` apparently is something distro-specific.  

Longer explanation:  The current Gentoo `locale-gen-3.x` is a Perl script that we can't utilize without updating our `dev-lang/perl`.  The previous `local-gen-2.y` from Gentoo is a bash script that probably will work as expected.  Gentoo seems to have borrowed the idea of `locale-gen` the script from Debian.  In other distros, the locales are distributed in binary form with `glibc` (or other package) and generating subsets of them on the running system seems not to be supported.

I'd like to continue supporting the use of specific locales.  Will try with Gentoo's `locale-gen-2.y` script.  It can be found here: https://github.com/gentoo/locale-gen/blob/locale-gen-2.23/locale-gen

security-kit/autogen.kit.d/security.yml
```diff
@@ -494,3 +494,78 @@ gnome_apps:
...
+        python? ( ${MY_PYTHON_DEPS} )
+      inherit:
+        - meson
+        - python-single-r1
+      iuse: caps man python seccomp test
+      python_compat: python3+
+      required_use: |
+        python? ( ${PYTHON_REQUIRED_USE} )
+        test? ( python )
+      restrict: !test? ( test )
```
geaaru: I think that you can drop `test` USE.

security-kit/autogen.kit.d/security.yml
```diff
@@ -494,3 +494,78 @@ gnome_apps:
...
+      license: GPL-2
+      extra_envs:
+        - |
+          MY_PYTHON_DEPS="
+            ${PYTHON_DEPS}
+            $(python_gen_cond_dep 'dev-python/pyelftools[${PYTHON_USEDEP}]')"
+      bdepend: |
+        caps? ( virtual/pkgconfig )
+        man? ( app-text/xmlto )
+        python? ( ${MY_PYTHON_DEPS} )
```
geaaru: you can drop this line from here considering that DEPEND use ${RDEPEND} the dependency is automatically used in build phase and so you can drop MY_PYTHON_DEPS and add directly the rows:
```
python? (
   ${PYTHON_DEPS}
   $(python_gen_cond_dep 'dev-python/pyelftools[${PYTHON_USEDEP}]')
)
```

security-kit/autogen.kit.d/security.yml
```diff
@@ -494,3 +494,78 @@ gnome_apps:
...
+            python-single-r1_pkg_setup
+          fi
+        }
+        src_configure() {
+          local emesonargs=(
+            "-Dlddtree_implementation=$(usex python python sh)"
+            $(meson_feature caps use_libcap)
+            $(meson_feature man build_manpages)
+            $(meson_use seccomp use_seccomp)
+            $(meson_use test tests)
```
geaaru: Removing `test` this will became `-Dtests=false`

security-kit/autogen.kit.d/security.yml
cuantar: Done!

security-kit/autogen.kit.d/security.yml
cuantar: Done!

security-kit/autogen.kit.d/security.yml
cuantar: Done!

cuantar:
Finished autogen for `locale-gen` based on Gentoo's github repo for the same script that we already have in our tree, as part of Gentoo's patches for our old `glibc` version.  It works as expected.  Alternatively, Debian's archive containing patches, etc. here: http://deb.debian.org/debian/pool/main/g/glibc/glibc_2.41-12+deb13u1.debian.tar.xz has a different implementation that we could use.

cuantar:
Local tests have succeeded!

core-kit/autogen.kit.d/templates/glibc.tmpl
```diff
@@ -0,0 +1,1397 @@
...
+
+inherit python-any-r1 preserve-libs toolchain-funcs flag-o-matic gnuconfig \
+	systemd multiprocessing tmpfiles
+
+DESCRIPTION="GNU libc C library"
+HOMEPAGE="https://www.gnu.org/software/libc/"
+
+KEYWORDS="*"
+SRC_URI="mirror://gnu/glibc/${P}.tar.xz"
+SRC_URI+=" https://dev.gentoo.org/~${PATCH_DEV}/distfiles/${P}-patches-${PATCH_VER}.tar.xz"
```
geaaru: I think that you need replace these with multiple assets in YAML. This will better support the Manifest generation too.
Something like this:
```yaml
vars:
   patch_version: '12'
   patch_dev: 'dilfridge'
   glibc_systemd_ver: xxx
assets:
  - name: '{{ .Values.pn }}-{{ .Values.version }}.tar.xz'
    url: 'https://ftp.gnu.org/gnu/glibc/{{ .Values.pn }}-{{ .Values.version }}.tar.xz'
  - name: '{{ .Values.pn }}-patches-{{ .Values.patch_version }}.tar.xz'
    url: 'https://dev.gentoo.org/~{{ .Values.patch_dev }}/distfiles/{{ .Values.pn }}-{{ .Values.version }}-patches-{{ .Values.patch_version }}.tar.xz'
  - name: '{{ .Values.pn }}-systemd-{{ .Values.glibc_systemd_ver }}.tar.gz'
    url: 'https://gitweb.gentoo.org/proj/toolchain/glibc-systemd.git/snapshot/{{ .Values.pn }}-systemd-{{ .Values.glibc_systemd_ver }}.tar.gz'
    use: systemd
```

cuantar: Done!  And a bit more, too.

cuantar: Still need to deal with the `*DEPEND` that's defined in the template instead of in the YAML.

cuantar: Had to figure out how to list multiple assets in the `SRC_URI`; found examples in other templates.  I think it's good now.

cuantar:
Again marked as draft until local tests pass.

geaaru:
Like this: https://github.com/macaroni-os/kit-fixups/blob/mark-unstable/templates/simple.tmpl#L25
or you can just use the `simple.tmpl` yaml directly ;)

cuantar:
Using the `simple.tmpl` YAML requires to move all of the ebuild logic into YAML, and it's a lot.  That's for the second pass ;) 

I did end up using that same code in the template.

cuantar:
There is some weird interaction between this generated ebuild and `libxcrypt` that has appeared, and I'm not sure why, yet.

cuantar:
I think it's all working as it should be.

The interaction with `libxcrypt` above has to do with the old `glibc-2.33` removing the `/lib64/libxcrypt.so.1` symlink and `/usr/include/crypt.h` header that it thinks it owns, after they're installed by `libxcrypt` as part of the upgrade procedure.  Manually re-creating the symlink allows `emerge` to operate again.  However!  in an empty `stage3` during a `mark-devkit metro` job, the task that installs `glibc` and friends (so, just `emerge glibc`) also picks up 26 additional packages.  One of those, `systemd-tmpfiles`, fails due to missing the header file.  The solution was to just let that task fail, and proceed to the next one (re-installing `libxcrypt`).  It is installed correctly during a future step for which it is a dep also.  In a live system performing a `glibc` upgrade, this problem with `systemd-tmpfiles` is not observed, because the latter package is already installed.

geaaru:
Good for a first integration in mark-unstable. Thanks!

PR #409, opened 2026-03-18
[kernel-kit] Autogen `sys-kernel/linux-headers` from kernel.org

cuantar:
This PR adds an autogen for the kernel headers from the vanilla source package at kernel.org.  

The `debian-sources` kernel package pulls sources from the same place.

I've left the minor version on it so that we know where it came from.  As a result, kernel headers will get updated over time, from whatever branch we choose.  Perhaps we should use a `SLOT` so that kernel autogens can depend on the correct headers and block headers that are newer than the kernel.

Note that the merge step is taken care of here, but will need to be adjusted for `mark-xl` and `mark-iii`, etc. because they're pinned to specific versions.

### DIFF
--- a/kernel-kit/autogen.kit.d/base.yml
+++ b/kernel-kit/autogen.kit.d/base.yml
@@ -1,4 +1,4 @@
-net_vpn_dirlisting:
+linux_kernel_dirlisting:
   generator: builtin-dirlisting
   template:
     engine: helm
@@ -90,3 +90,98 @@ net_vpn_dirlisting:
               # Don't forget to umount /boot if it was previously mounted by us.
               use initramfs && ego_pkg_postrm
             }
+
+    - vanilla-headers:
+        dir:
+          url: https://mirrors.edge.kernel.org/pub/linux/kernel/v6.x/
+          matcher: 'linux-.*.tar.xz'
+        selector:
+          - '>6.12'
+          - '<6.13'
+        transform:
+          - kind: string
+            match: 'linux-'
+            replace: ''
+          - kind: string
+            match: '.tar.xz'
+            replace: ''
+        vars:
+          category: sys-kernel
+          bdepend: |
+            app-arch/xz-utils
+            dev-lang/perl
+            net-misc/rsync
+            !<sys-kernel/linux-headers-6.12
+            !!sys-kernel/debian-headers
+          desc: Linux kernel header files
+          homepage: https://kernel.org
+          inherit:
+            - eutils
+            - toolchain-funcs
+          license: GPL-2
+          patches:
+            - 00_all_0002-x86-do-not-build-relocs-tool-when-installing-headers.patch
+          restrict: binchecks strip
+          s: '${WORKDIR}/linux-${PV}'
+          body: |
+            pkg_setup() {
+              export REAL_ARCH="$ARCH"
+              # will interfere with Makefile if set
+              unset ARCH; unset LDFLAGS
+            }
+            src_prepare() {
+              default
+              # drop unused errno.h include
+              sed -e '/^#include <errno.h>/d' -i ${WORKDIR}/scripts/unifdef.c
+            }
+            src_compile() {
+              :
+            }
+            src_install() {
+              # fix permissions in source tree
+              cd "${WORKDIR}"
+              chown -R 0:0 * >& /dev/null
+              chmod -R a+r-w+X,u+w *
+              cd "${OLDPWD}"
+              [[ -L /usr/include/linux ]] && { rm /usr/include/linux || die; }
+              [[ -L /usr/include/asm ]] && { rm /usr/include/asm || die; }
+              emake O="${D}" ARCH="$(tc-arch-kernel)" headers_install \
+                || die "could not install headers to image"
+              # prune non-headers from tree
+              rm "${D}"/{arch,include,scripts} -r
+              find "${D}" '(' -name '.install' -o -name '*.cmd' ')' -delete
+            }
+
+virtual_os_headers:
+  generator: builtin-noop
+  defaults:
+    template: ../../templates/simple.tmpl
+    files_dir: "{{ .Values.pn }}"
+    ignore_artefacts: true
+  packages:
+    - linux-headers:
+        category: sys-kernel
+        vars:
+          depend: ${RDEPEND}
+          desc: "Virtual for Linux kernel headers"
+          homepage: https://kernel.org
+          license: GPL-2
+          rdepend: |
+            || (
+              >=sys-kernel/debian-headers-{{ .Values.version }}
+              >=sys-kernel/vanilla-headers-{{ .Values.version }}
+            )
+            !<sys-kernel/linux-headers-{{ addf .Values.version 0.01 }}
+            !>sys-kernel/debian-headers-{{ addf .Values.version 0.01 }}
+            !>sys-kernel/vanilla-headers-{{ addf .Values.version 0.01 }}
+          versions:
+            - '6.12'
+    - os-headers:
+        category: virtual
+        vars:
+          desc: "Virtual for operating system headers"
+          # Presently we only support Linux
+          depend: ${RDEPEND}
+          rdepend: sys-kernel/linux-headers
+          versions:
+            - '1'
--- a/kernel-kit/autogen.kit.d/debian.yml
+++ b/kernel-kit/autogen.kit.d/debian.yml
@@ -100,6 +100,81 @@ debian_sources:
             - 6.1.79+/more-ISA-levels-and-uarches-for-kernel-6.1.79+.patch
           config_extract: 6.1
 
+debian_headers:
+  generator: custom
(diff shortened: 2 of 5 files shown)


geaaru:
What do you think to uses the name `sys-kernel/debian-headers` and autogen `virtual/os-headers`?
In order to have something like this:
```
RDEPEND="
    || (
         sys-kernel/debian-headers
         sys-kernel/vanilla-headers
    )
"
DEPEND="${RDEPEND}"
```

The use of one or other could be later controlled from a mix-in.

geaaru:
About the SLOT i think that it's more easy to manage always SLOT 0 and mask packages... Else it's a pain to manage upgrades.

geaaru:
Sorry, i read better the PR now and i see that is based already on vanilla. So, i mean the inverse... for this `vanilla-headers` and to add a `debian-headers` too? Do you think that could be fine to have a patched headers file too?

cuantar:
> Sorry, i read better the PR now and i see that is based already on vanilla. So, i mean the inverse... for this `vanilla-headers` and to add a `debian-headers` too? Do you think that could be fine to have a patched headers file too?

Yes; will work on a patched `debian-headers` too, using the generator that fetches sources, patches, and configs for the `debian-sources` kernel.  For that generator, the sources come from kernel.org, and the patches & configs are from a Debian repo mirror.

cuantar:
Here's a working `debian-headers`, too.

The logic is:
- `virtual/os-headers` has been simplified only to depend on `sys-kernel/linux-headers`;
- `sys-kernel/linux-headers` is now a virtual package with `major.minor` version like `linux-headers-6.12`; it depends on either `sys-kernel/vanilla-headers` or `sys-kernel/debian-headers` with the same version pair, but patchlevel is allowed to float;
- `sys-kernel/vanilla-headers` is purely from upstream, with two trivial patches to break a circular dependency and let it build before the headers are installed; and
- `sys-kernel/debian-headers` is `vanilla-headers` with Debian patches, as with the `debian-sources`.

There are >90 packages that depend explicitly on `sys-kernel/linux-headers` and check its version; and ~ 35 that depend on `virtual/os-headers`.

cuantar:
The upgrade procedure is:
- `emerge -C linux-headers`
- `emerge -1v os-headers linux-headers vanilla-headers` (or `debian-headers`, by your preference)

If `linux-headers` is not removed first, there will be a file collision when installing the `*-headers` packages that have different names.

cuantar:
Closes macaroni-os/mark-issues#92.

kernel-kit/autogen.kit.d/base.yml
```diff
@@ -90,3 +90,98 @@ net_vpn_dirlisting:
...
+                || die "could not install headers to image"
+              # prune non-headers from tree
+              rm "${D}"/{arch,include,scripts} -r
+              find "${D}" '(' -name '.install' -o -name '*.cmd' ')' -delete
+            }
+
+virtual_os_headers:
+  generator: builtin-noop
+  template:
+    engine: helm
```
geaaru: This could be omitted, it's the default

kernel-kit/autogen.kit.d/vanilla-headers/00_all_0001-unifdef-drop-unused-errno.h-include.patch
```diff
@@ -0,0 +1,32 @@
...
+
+diff --git a/scripts/unifdef.c b/scripts/unifdef.c
+index 7493c0ee51cc..c5dfae538b08 100644
+--- a/scripts/unifdef.c
++++ b/scripts/unifdef.c
+@@ -48,7 +48,6 @@
+ 
+ #include <ctype.h>
+ #include <err.h>
+-#include <errno.h>
```
geaaru: We can convert this is a sed probably in the future

kernel-kit/autogen.kit.d/base.yml
```diff
@@ -90,3 +90,98 @@ net_vpn_dirlisting:
...
+            pkg_setup() {
+              export REAL_ARCH="$ARCH"
+              # will interfere with Makefile if set
+              unset ARCH; unset LDFLAGS
+            }
+            src_compile() {
+              :
+            }
+            src_test() {
+              emake ${MAKEOPTS} O="${WORKDIR}"/build headers_check || die "check failed"
```
geaaru: i'm not sure that we need `src_test`

kernel-kit/autogen.kit.d/base.yml
```diff
@@ -90,3 +90,98 @@ net_vpn_dirlisting:
...
+          desc: "Virtual for Linux kernel headers"
+          homepage: https://kernel.org
+          license: GPL-2
+          rdepend: |
+            || (
+            >=sys-kernel/debian-headers-{{ .Values.version }}
+            >=sys-kernel/vanilla-headers-{{ .Values.version }}
+            )
+            !>sys-kernel/debian-headers-{{ addf .Values.version 0.01 }}
+            !>sys-kernel/vanilla-headers-{{ addf .Values.version 0.01 }}
```
geaaru, line 176: Could be correct to set !<sys-kernel/linux-headers-<current-version> to force uninstall?

kernel-kit/autogen.kit.d/debian.yml
```diff
@@ -100,6 +100,85 @@ debian_sources:
...
+            eapply -p1 "${WORKDIR}/debian/patches/${debpatch}" || die "could not apply patch!"
+          done
+        }
+        src_compile() {
+          :
+          #install -d "${WORKDIR}"/build || die "problem making build dir"
+          #emake O="${WORKDIR}"/build headers || die "could not build header tree"
+        }
+        src_test() {
+          emake O="${WORKDIR}"/build headers_check || die "check failed"
```
geaaru: Same here, i don't see `IUSE=test` so probably this could be removed.

kernel-kit/merge.kit.d/base.yml
```diff
@@ -33,6 +33,11 @@ target:
...
 
   atoms:
+    - pkg: sys-kernel/linux-headers
+      conditions:
+        - '>=sys-kernel/linux-headers-6.12'
+        - '<sys-kernel/linux-headers-6.13'
+
     - pkg: sys-kernel/linux-headers
       conditions:
         - '>=sys-kernel/linux-headers-6.9'
```
geaaru: i dunno if here we need <6.12

geaaru:
Excluding optional fix described in my comments the PR seems OK. Feel free to merge it if i'm not around.

cuantar:
I will fix the things you mentioned in your comments.  It all makes sense.

kernel-kit/merge.kit.d/base.yml
cuantar: Done!

kernel-kit/autogen.kit.d/debian.yml
cuantar: Done!

kernel-kit/autogen.kit.d/base.yml
cuantar: Hm, done, but may need to be tweaked with more testing.  Good enough for now!

kernel-kit/autogen.kit.d/base.yml
cuantar: Done!

kernel-kit/autogen.kit.d/base.yml
cuantar: Done!

kernel-kit/autogen.kit.d/vanilla-headers/00_all_0001-unifdef-drop-unused-errno.h-include.patch
cuantar: Done, but needs testing before this PR gets merged.

kernel-kit/autogen.kit.d/base.yml
```diff
@@ -90,3 +90,98 @@ net_vpn_dirlisting:
...
+          body: |
+            pkg_setup() {
+              export REAL_ARCH="$ARCH"
+              # will interfere with Makefile if set
+              unset ARCH; unset LDFLAGS
+            }
+            src_prepare() {
+              default
+              # drop unused errno.h include
+              sed -Ee 's/^#include <errno.h>//g' -i ${WORKDIR}/scripts/unifdef.c
```
geaaru: in alternative you can drop the line directly. It seems more safe.

```
sed -e '/^#include <errno.h>/d' -i ${WORKDIR}/scripts/unifdef.c
```

cuantar: okay, agreed, that does seem a bit cleaner.  will change it.

cuantar:
I think this PR is ready now!

geaaru:
It sounds good. Another little step for our autogen and one giant leap for Macaroni OS!
Thank you very much for all this work.

PR #410, opened 2026-04-10
Update bind-tools-21.tmpl

coffnix:
bind-tools requires the dev-db/lmdb ebuild to compile.


cuantar:
Looks good! Thanks for fixing it.

PR #411, opened 2026-04-10
Update bind-21.tmpl

coffnix:
lmdb is no longer optional, is mandatory


cuantar:
See https://lists.isc.org/pipermail/bind-users/2026-March/110528.html

PR #416, opened 2026-05-08
profiles: add banana-pi-a7 + banana-pro-a20

coffnix:
Adds two new ARM 32-bit Banana Pi subarch profiles using CHOST=armv7a-unknown-linux-gnueabihf for compatibility with VFScratch and modern GCC bootstrap toolchains.

The new banana-pi-cortex-a7 profile provides a conservative and generic Cortex-A7 hardfloat target using -mfpu=vfpv3-d16, intended to work across multiple older Banana Pi ARMv7 boards without assuming NEON or VFPv4 support.

The new banana-pro-a20 profile provides a more optimized target specifically for Allwinner A20 based Banana Pro boards, enabling -mfpu=neon-vfpv4 and NEON related CPU flags for better performance on hardware known to support these extensions.

The existing ARMv7 hardfp profiles use armv7a-hardfloat-linux-gnueabi, which does not match the triplet used by VFScratch in this environment. This can cause Portage to select toolchain prefixes that do not exist during bootstrap stages.

These new profiles avoid CHOST/toolchain prefix mismatches while also providing clearer separation between generic ARMv7 compatibility targets and board-specific optimized profiles.


geaaru:
Thanks!

PR #417, opened 2026-05-10
[python-kit] Update body placement in autogen script. it should be under vars

org-tekeli-borisp:
Related to github.com/macaroni-os/mark-issues/issues/673
Thanks to @cuantar!


org-tekeli-borisp:
Tested
image
image

PR #419, opened 2026-05-13
[net-kit] Autogen net-firewall/iptables

cuantar:
It merges!  But I haven't tested it in a live system yet.

### DIFF
--- a/net-kit/autogen.kit.d/base.yml
+++ b/net-kit/autogen.kit.d/base.yml
@@ -3318,3 +3318,151 @@ gitlab_various:
             src_install() {
                 emake INSTALL_ROOT="${D}" install
             }
+
+
+iptables_dirlisting:
+  generator: builtin-dirlisting
+
+  defaults:
+    dir:
+      url: https://www.netfilter.org/pub/iptables
+      matcher: "iptables-.*.tar.xz$"
+    assets:
+      - name: "{{ .Values.pn }}-{{ .Values.version }}.tar.xz"
+    transform:
+      - kind: string
+        match: 'iptables-'
+        replace: ''
+      - kind: string
+        match: '.tar.xz'
+        replace: ''
+    files_dir: "{{ .Values.pn }}"
+    template: ../../templates/simple.tmpl
+    category: net-firewall
+    match:
+
+    vars:
+      desc: 'Linux kernel (2.4+) firewall, NAT and packet mangling tools'
+      homepage: https://www.netfilter.org/projects/iptables/
+      iuse: conntrack ipv6 netlink nftables pcap static-libs
+      license: GPL-2
+
+      bdepend: |-
+        ${CDEPEND}
+        virtual/pkgconfig
+        nftables? (
+          sys-devel/flex
+          virtual/yacc
+        )
+      cdepend: |-
+        conntrack? ( net-libs/libnetfilter_conntrack )
+        netlink? ( net-libs/libnfnetlink )
+        nftables? (
+          net-libs/libmnl:=
+          net-libs/libnftnl:=
+        )
+        pcap? ( net-libs/libpcap )
+
+      depend: |-
+        ${CDEPEND}
+        sys-kernel/linux-headers
+        virtual/os-headers
+
+      rdepend: |-
+        ${CDEPEND}
+        nftables? ( net-misc/ethertypes )
+
+      inherit:
+        - autotools
+        - flag-o-matic
+        - toolchain-funcs
+
+      body: |-
+        src_prepare() {
+          # use the saner headers from the kernel
+          rm include/linux/{kernel,types}.h || die
+
+          # Only run autotools if user patched something
+          eapply_user && eautoreconf || elibtoolize
+        }
+
+        src_configure() {
+          # Some libs use $(AR) rather than libtool to build #444282
+          tc-export AR
+
+          # Hack around struct mismatches between userland & kernel for some ABIs. #472388
+          use amd64 && [[ ${ABI} == "x32" ]] && append-flags -fpack-struct
+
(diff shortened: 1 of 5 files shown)


net-kit/autogen.kit.d/base.yml
```diff
@@ -3318,3 +3318,155 @@ gitlab_various:
...
+    files_dir: "{{ .Values.pn }}"
+    template: ../../templates/simple.tmpl
+    category: net-firewall
+    match:
+
+    vars:
+      desc: 'Linux kernel (2.4+) firewall, NAT and packet mangling tools'
+      extra_envs:
+        - |-
+          COMMON_DEPEND="
```
geaaru: It sounds like an AI generation issue 😱 
We have a `cdepend` variable for this that could be later used with `${CDEPEND}`

net-kit/autogen.kit.d/base.yml
```diff
@@ -3318,3 +3318,155 @@ gitlab_various:
...
+      bdepend: |-
+        ${COMMON_DEPEND}
+        virtual/pkgconfig
+        nftables? (
+          sys-devel/flex
+          virtual/yacc
+        )
+
+      depend: |-
+        ${COMMON_DEPEND}
```
geaaru: Like shared before this must be converter in `${CDEPEND}`

net-kit/autogen.kit.d/base.yml
```diff
@@ -3318,3 +3318,155 @@ gitlab_various:
...
+        sys-kernel/linux-headers
+        virtual/os-headers
+
+      rdepend: |-
+        ${COMMON_DEPEND}
```
geaaru: Like shared before this must be converter in `${CDEPEND}`

net-kit/autogen.kit.d/base.yml
```diff
@@ -3318,3 +3318,155 @@ gitlab_various:
...
+            $(use_enable pcap bpf-compiler) # W: Prefer mapfile or read -a to split command output (or quote to avoi…
+            $(use_enable pcap nfsynproxy) # W: Prefer mapfile or read -a to split command output (or quote to avoid …
+            $(use_enable static-libs static) # W: Prefer mapfile or read -a to split command output (or quote to avo…
+            $(use_enable ipv6) # W: Prefer mapfile or read -a to split command output (or quote to avoid splitting).
+          )
+          econf "${myeconfargs[@]}"
+        }
+
+        src_compile() {
+          emake V=1
```
geaaru: Is this really needed?

net-kit/autogen.kit.d/base.yml
```diff
@@ -3318,3 +3318,155 @@ gitlab_various:
...
+          rm -f "${ED}"/sbin/iptables
+          rm -f "${ED}"/sbin/ip6tables
+          rm -f "${ED}"/sbin/iptables-save
+          rm -f "${ED}"/sbin/iptables-restore
+          rm -f "${ED}"/sbin/ip6tables-save
+          rm -f "${ED}"/sbin/ip6tables-restore
+
+          if use nftables; then
+            rm "${ED}"/etc/ethertypes || die
+            rm "${ED}"/sbin/{arptables,ebtables}{,-{save,restore}} || die
```
geaaru: Considering that we have `arptables` and `ebtables` packages this must be done in any case i think.
Same for `ethertypes`.
Second point is: i'm not sure of this... when i created the new arptables and ebtables autogen i saw that they are now called `arptables-legacy` and `ebtables-legacy`. So the question is: does these binaries to install? 🤔

geaaru (changes requested):

net-kit/autogen.kit.d/base.yml
cuantar: Haha no :D  I wrote this one by hand!
I just didn't know about our `cdepend` variable.  That's much nicer; will fix!

net-kit/autogen.kit.d/base.yml
cuantar: maybe not, will experiment.  it's from our ebuild

net-kit/autogen.kit.d/base.yml
cuantar: Good question.  There are no symlinks made for those.  Maybe there should be.

geaaru: i'm not sure of this. I don't think that they are used from particular tools. It's been sufficient update init.d script correctly to continue using them like before and so in this case could be fine probably just having the right name without create tons of symlinks everywhere.

net-kit/autogen.kit.d/base.yml
geaaru: It sounds like a cut&paste from history not really needed. I think that the `src_compile` could be dropped

geaaru:
I think that there is also another issue. It's missed the `inherit` variable.

net-kit/autogen.kit.d/base.yml
cuantar: yes, verified:  correct, it's unnecessary.  dropped.

net-kit/autogen.kit.d/base.yml
cuantar: I moved the `rm` calls above the `nftables` test.  Now only the ebuilds for those packages will install binaries for them.

geaaru:
@cuantar it seems still missed the `inherit` variable.
Second point: why we need to keep in the PRs all old *init.d/conf.d* files? We don't need them in kit-fixups, just keep the rights files there so in the near future will be more easy clean files not present there.

cuantar:
@geaaru I think it's good now:  old files removed, and `inherit` added where it should be in the spec.

PR #424, opened 2026-05-17
[kde-kit] Autogen libqaccessibilityclient

linuxexplorer:
Autogen added with qt5 and qt6 use flags to allow backwards compatibility for now
[closes]: macaroni-os/mark-issues#682

### DIFF
--- a/kde-kit/autogen.kit.d/base.yml
+++ b/kde-kit/autogen.kit.d/base.yml
@@ -790,3 +790,55 @@ calamares:
 
   packages:
     - calamares:
+
+libqaccessibilityclient:
+  generator: builtin-dirlisting
+  template:
+    engine: helm
+
+  defaults:
+    files_dir: "{{ .Values.pn }}"
+    template: ../../templates/simple.tmpl
+    category: media-libs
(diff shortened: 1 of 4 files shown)


kde-kit/autogen.kit.d/base.yml
```diff
@@ -790,3 +790,55 @@ calamares:
...
+      url: 'https://download.kde.org/Attic/libqaccessibilityclient/'
+      matcher: "libqaccessibilityclient-[0-9.]+\\.tar\\.xz$"
+    transform:
+      - kind: string
+        match: 'libqaccessibilityclient'
+        replace: ''
+      - kind: string
+        match: '.tar.xz'
+        replace: ''
+    assets:
```
geaaru, line 813: Probably this is not needed because the assets is the same used for the retrieve the available versions but move ahead in this way for now.

geaaru:
I dunno if could be a problem to change the SLOT and replace the existing for the qt5 flag. Probably could be better add a conflicts to the previous version but considering that i'm trying to dismiss qt5 this has less priority.

PR #428, opened 2026-05-25
[dev-kit] Fix dev-libs/nss postinst fails on ARM32 systems due to har…

coffnix:
…dcoded /usr/lib64/libfreebl3.so path

This PR improves error handling for NSS whip hooks during postinst and postrm phases.

Current implementation does not explicitly abort the ebuild when the whip helper fails, which may leave the system with incomplete NSS hook execution or partially configured libraries.

The updated implementation adds proper failure checks using die. Reported on https://github.com/macaroni-os/mark-issues/issues/695

This ensures that hook failures are immediately visible and prevents silent or inconsistent NSS installation states.

### DIFF
--- a/dev-kit/autogen.kit.d/base.yml
+++ b/dev-kit/autogen.kit.d/base.yml
@@ -512,10 +512,10 @@ linear_packages:
                 > "${ED}"/etc/prelink.conf.d/nss.conf
             }
             pkg_postinst() {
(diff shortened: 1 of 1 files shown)


geaaru:
Thanks

coffnix:
Tested successfully on the following architectures:

- ARM32
- ARM64
- x86_64

PR #430, opened 2026-05-26
[net-kit] net-firewall/iptables-1.8.13: fix install failure when USE=…

coffnix:
…”-nftables”

iptables: avoid removing nftables-only files when USE=”-nftables”

When USE=”-nftables”, some nftables-related files and binaries are not installed by upstream anymore, causing the ebuild to fail during src_install() while trying to remove non-existent files such as /etc/ethertypes and arptables/ebtables nft wrappers.

This patch adds proper USE conditional checks so these removals only happen when USE=“nftables” is enabled, matching the actual installed file set and preventing installation failures on legacy-only builds.

Reported on: https://github.com/macaroni-os/mark-issues/issues/697

### DIFF
--- a/net-kit/autogen.kit.d/base.yml
+++ b/net-kit/autogen.kit.d/base.yml
@@ -3442,8 +3442,8 @@ iptables_dirlisting:
           rm -f "${ED}"/sbin/iptables-restore
           rm -f "${ED}"/sbin/ip6tables-save
           rm -f "${ED}"/sbin/ip6tables-restore
-          rm "${ED}"/etc/ethertypes || die
-          rm "${ED}"/sbin/{arptables,ebtables}{,-{save,restore}} || die
+          rm "${ED}"/etc/ethertypes || ewarn "File /etc/ethertypes not present. Ignoring."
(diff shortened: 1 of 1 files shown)


net-kit/autogen.kit.d/base.yml
```diff
@@ -3442,8 +3442,10 @@ iptables_dirlisting:
           rm -f "${ED}"/sbin/iptables-restore
           rm -f "${ED}"/sbin/ip6tables-save
           rm -f "${ED}"/sbin/ip6tables-restore
-          rm "${ED}"/etc/ethertypes || die
```
geaaru, line 3445: Why just convert these in `|| true`  or `|| { ewarn "File ethertypes not present. Ignoring" }` ?
Considering that we have few time to review every time every single ebuild, when there are files removed from the installation and these files are not present we can add a warning or just ignore it.
The result is the same, We want remove a file and if the file is no more present the result is the same.
Can you review this patch?

geaaru (changes requested):

coffnix (commented):
please check now @geaaru

PR #431, opened 2026-05-26
Update dracut-mark.conf

coffnix:
This is just a small fix to adjust the path to the correct location: /usr/lib64/elogind/elogind-uaccess-command.

### DIFF
--- a/kernel-kit/autogen.kit.d/virtual/dracut-mark/dracut-mark.conf
+++ b/kernel-kit/autogen.kit.d/virtual/dracut-mark/dracut-mark.conf
@@ -1,8 +1,10 @@
 # Macaroni OS (MARK) Dracut Default parameters
 
-dracutmodules="rescue resume bash kernel-modules rootfs-block udev-rules usrmount base fs-lib shutdown terminfo shutdown"
+dracutmodules="rescue resume bash kernel-modules rootfs-block udev-rules usrmount base fs-lib shutdown terminfo crypt lvm mdraid shutdown"
+
+add_drivers+=" dm_mod dm_crypt "
 
 omit_dracutmodules="fcoe"
 
-# Quiet messages: failed to execute '/lib64/elogind/elogind-uaccess-command' '/lib64/elogind/elogind-uaccess-command /dev/dri/card0 ': No such file or directory
-install_items="/lib64/elogind/elogind-uaccess-command"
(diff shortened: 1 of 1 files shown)


geaaru:
About your discord question for `$(get_libdir)` probably you can do this:
- in the `dracut-mark.conf` you can use a placeholder like this:
```
install_items="LIBDIR/elogind/elogind-uaccess-command"
```
- and later in the ebuild add something like this:
```
sed -i -e "s|LIBDIR|/usr/$(get_libdir)|g" ${ED}/etc/dracut.conf.d/99-macaroni.conf
```
This will support both 64 bit and 32 bit.

Btw, i have a doubt, atm in Phoenix i have this:
```
$ anise q files elogind | grep uaccess
/lib/udev/rules.d/70-uaccess.rules
/lib64/elogind/elogind-uaccess-command
```
The path is without `/usr`. What version do you use?

coffnix:
my elogind on overlay local have ```--libexecdir="$(get_libdir)/elogind"```

geaaru:
So the path is `/lib` or `/lib64` not `/usr/lib|/usr/lib64`.
The fix in the PR is not valid

geaaru:
I dunno if there is a `--prefix=/usr` that is used as prefix

geaaru:
So, there are tree things to do:
1) Fix our autogen: https://github.com/macaroni-os/kit-fixups/blob/mark-unstable/security-kit/autogen.kit.d/security.yml#L198 in order to use `$(get_libdir)`
2) Set this on dracut-mark.conf: `install_items="/usr/LIBDIR/elogind/elogind-uaccess-command"`
3) Add something like this:
```
sed -i -e "s|LIBDIR|$(get_libdir)|g" ${ED}/etc/dracut.conf.d/99-macaroni.conf
```
here https://github.com/macaroni-os/kit-fixups/blob/mark-unstable/kernel-kit/autogen.kit.d/debian.yml#L198

coffnix:
fixed @geaaru I also added the crypt and lvm dracut modules to the default configuration. Without these modules, systems using a LUKS-encrypted root filesystem with LVM are unable to boot correctly because the initramfs cannot unlock the encrypted device and activate the volume groups during the early boot stage.

Additionally, the required elogind helper binary is now installed into the initramfs to ensure proper device access handling during early userspace initialization.

coffnix:
included mdraid module, thanks @geaaru

PR #432, opened 2026-06-02
Fix libxml2-python removing Python bindings during install

coffnix:
The previous install cleanup used:
```
rm -r "${ED}"/usr/{$(get_libdir),bin,include} || die
```

This was too aggressive because it removed the entire ${ED}/usr/lib directory. As a result, the Python bindings installed under site-packages were deleted during src_install(), causing import libxml2 to fail even though the package built successfully.

This patch replaces the broad directory removal with targeted cleanup of only the files already provided by dev-libs/libxml2, preserving the Python module and its installation path while still avoiding file collisions with the main libxml2 package.

### DIFF
--- a/python-modules-kit/autogen.kit.d/python.yml
+++ b/python-modules-kit/autogen.kit.d/python.yml
@@ -892,7 +892,13 @@ libxml2_python:
         src_install() {
           meson_src_install
           # Remove files supplied by libxml2 library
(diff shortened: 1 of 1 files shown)


coffnix:
As suggested by @geaaru , it may be better to create a separate branch for x86_32bit.

coffnix:
It is better to keep:
```
rm -r "${ED}"/usr/{lib64,bin,include} || die
```

for 64-bit systems @geaaru

PR #433, opened 2026-06-20
Initial xfce autogens

linuxexplorer:
Initial xfce autogens, for current stable 4.20 series

4.22 series will use meson build and the autogen will need to change to reflect this


geaaru:
Thank you very much! :rocket:

PR #436, opened 2026-08-08
Issue-717:obs-studio does not merge with Qt6

linuxexplorer:
Autogen obs-studio and missing deps.
Tested and merges successfully.

Initial commit - needs testing.
Missing deps have been added to media-kit autogen but could be a better fit in a different kit.
Some ebuild contents may not fit the mark conventions - let me know what needs to be changed ;-)
I've not included a fallback qt5 version - assumed we're moving forward with Qt6 versions only.

### DIFF
--- a/media-kit/autogen.kit.d/base.yml
+++ b/media-kit/autogen.kit.d/base.yml
@@ -4754,14 +4754,13 @@ obs_studio:
             match: 'OBS Studio '
             replace: ''
         assets:
-          - name: "{{ .Values.pn }}-{{ .Values.version }}.tar.gz"
-            url: "https://api.github.com/repos/{{ .Values.github_user }}/{{ .Values.github_repo }}/tarball/{{ .Values.version }}"
-          - name: "{{ .Values.cef_tarball }}.tar.xz"
-            url: "https://cdn-fastly.obsproject.com/downloads/{{ .Values.cef_tarball }}.tar.xz"
+          - name: '{{ .Values.pn }}-{{ .Values.version }}.tar.gz'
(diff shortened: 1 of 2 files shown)


media-kit/autogen.kit.d/base.yml
```diff
@@ -4728,3 +4728,511 @@ flite:
...
+            replace: ''
+        assets:
+          - name: "{{ .Values.pn }}-{{ .Values.version }}.tar.gz"
+            url: "https://api.github.com/repos/{{ .Values.github_user }}/{{ .Values.github_repo }}/tarball/{{ .Values.version }}"
+          - name: "{{ .Values.cef_tarball }}.tar.xz"
+            url: "https://cdn-fastly.obsproject.com/downloads/{{ .Values.cef_tarball }}.tar.xz"
+            use: browser
+        vars:
+          cef_dir: cef_binary_6533_linux_x86_64
+          cef_tarball: cef_binary_6533_linux_x86_64_v6
```
geaaru, line 4764: It sounds better just keep the version as var with value `6533` and move the tarball syntax over the assets section. Will be more clear for user if there is an update to do.
Something like this:
```
vars:
   cef_version: '6533'
assets:
  - name: '{{ .Values.pn }}-cef-binary-{{ .Values.cef_version }}-x86-64_v6.tar.xz'
    url: 'https://cdn-fastly.obsproject.com/downloads/cef_binary_{{ .Values.cef_version }}_linux_x86_64_v6.tar.xz'
```
I added also on `name` the `PN` prefix considering that this package is strictly related to the obs-studio package.

geaaru:
Thank you very much :rocket:

geaaru:
Related to https://github.com/macaroni-os/mark-issues/issues/717

PR #437, opened 2026-08-09
Issue 707: Add QT6 support to lightdm

linuxexplorer:
Initial autogen for lightdm - needs testing

### DIFF
--- a/desktop-kit/autogen.kit.d/base.yml
+++ b/desktop-kit/autogen.kit.d/base.yml
@@ -646,6 +646,196 @@ general_github:
 
   packages:
 
+    - lightdm:
+        category: x11-misc
+        github:
+          user: canonical
+          query: tags
+        vars:
+          homepage: https://github.com/canonical/lightdm
+          desc: A lightweight display manager
+          license: GPL-3 LGPL-3
+          inherit:
+            - autotools
+            - flag-o-matic
+            - pam
+            - qmake-utils
+            - systemd
+            - user
+            - vala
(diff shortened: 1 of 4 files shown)


desktop-kit/autogen.kit.d/base.yml
```diff
@@ -646,6 +646,196 @@ general_github:
...
+            vala? ( $(vala_depend) )
+          pdepend: |
+            gtk? ( x11-misc/lightdm-gtk-greeter )
+          body: |
+            pkg_setup() {
+              export LIGHTDM_USER=${LIGHTDM_USER:-lightdm}
+              if use non_root ; then
+                enewgroup ${LIGHTDM_USER}
+                enewgroup video
+                enewuser ${LIGHTDM_USER} -1 -1 /var/lib/${LIGHTDM_USER} ${LIGHTDM_USER},video
```
geaaru: It could be fine instead of using dynamic user/group to use a static id.

https://github.com/macaroni-os/macaroni-commons/blob/master/packages/entities/groups.yaml#L1124

In `entities-catalog` is mapped as 134 uid/gid.

linuxexplorer: Thanks, I've updated the PR but not sure if I've done it correctly

geaaru:
Thank you very much! :rocket:

geaaru:
Related to https://github.com/macaroni-os/mark-issues/issues/704

PR #438, opened 2026-08-10
Issue 705: matplotlib needs a qt6 use

linuxexplorer:
Initial matplotlib autogen + deps
All pinned due to python version required
Needs testing


geaaru:
Related to https://github.com/macaroni-os/mark-issues/issues/705

Thanks! :rocket:

PR #439, opened 2026-08-10
Issue 735: autogen gpgme and associated packages to enable qt6

linuxexplorer:
Initial autogens
This may break existing deps elsewhere - would need to be fixed or older version retained until then

### DIFF
--- a/security-kit/autogen.kit.d/security.yml
+++ b/security-kit/autogen.kit.d/security.yml
@@ -675,3 +675,214 @@ xmlsec:
         selector:
           - '>=1.3.0'
 
+gnupg_libs:
+  generator: builtin-dirlisting
+  generator_opts:
+    rate_limiter: "1"
+
+  defaults:
+    category: app-crypt
+    files_dir: "{{ .Values.pn }}"
+    template: ../../templates/simple.tmpl
+    dir:
+      url: https://gnupg.org/ftp/gcrypt/{{ .Values.pn }}/
+      exclude: 'rc|beta|alpha'
+    vars:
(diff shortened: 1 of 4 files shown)


security-kit/merge.kit.d/security_autogen.yml
```diff
@@ -27,8 +27,22 @@ target:
       conditions:
         - '<app-crypt/gcr-4.0'
 
+    - pkg: app-crypt/gpgme
+      conditions:
```
geaaru: FYI: The use of the `conditions` is needed only when:
a) we want keep an old version available and avoid that the bot remove it in the clean task
b) we have multiple version and we want keep the three versions for a specific rance (for example nodjes 24.x, 22.x, etc.)

I'm not sure that in this case we want to keep old version of the `gpgme` package

security-kit/autogen.kit.d/security.yml
```diff
@@ -675,3 +675,214 @@ xmlsec:
...
+              local mycmakeargs=(
+                -DBUILD_TESTING=OFF
+              )
+
+              cmake_src_configure
+            }
+
+    - qgpgme:
+        selector:
+          - '>=2.1.0'
```
geaaru: By default `mark-devkit` take the last available version so the selector is needed normally when we want pin a previous version with `<=` or again for example for packages with multiples releases like nodejs where we want: `<25.0`, `>24.0`
Is there a reason for this selector that you have saw with `doit` ?

linuxexplorer: I've removed the conditions and selectors

PR #444, opened 2026-08-13
[core-kit] Autogen `sys-devel/gcc`

cuantar:
A custom-extension determines the version of the math libraries to use.  Gentoo patches are applied by default, but are gated behind `USE=-vanilla` that can be changed to skip them.

Note that the latest version in the `14.x` branch is `gcc-14.4.0`.  I think we should target that version first.

The autogen will generate `gcc-{14..16}.x.0`.


cuantar:
As an experiment, I used a local agent backed by an LLM running on local hardware to develop this autogen.  The spec is not "vibe-coded" --- the model generated it over many iterations with my guidance, and with a testing loop.  I've tested the result locally and each version builds.

cuantar:
I've now noticed that `USE=stack_clash_protection` that used to be turned on is missing.  Need to fix.

cuantar:
Also note:  It's intentional to be using the `gcc-14.3.0` patches with `gcc-14.4`.  Gentoo does it that way, too.  I commented the ones that don't apply from the variable in the spec.  If we want to pin `gcc-14.3.0` then we should fix those.

geaaru:
Add reference to related issue https://github.com/macaroni-os/mark-issues/issues/612

geaaru:
Could be good create a blog post in our website that describe how you have did this autogen ;)

Thank you very much!! :rocket: :tada:

PR #445, opened 2026-08-14
[templates] Fix tabs in ebuild body, in simple.tmpl

cuantar:
Here is a local fix I've been using for a while (some months) to the formatting in ebuilds resulting from `simple.tmpl`.  The robot figured it out while I did something else.  This template change improves nested indentation handling.

### DIFF
--- a/templates/simple.tmpl
+++ b/templates/simple.tmpl
@@ -94,7 +94,17 @@ S="{{ .Values.s }}"
 {{- end }}
 
 {{- if .Values.body }}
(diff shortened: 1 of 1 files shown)


geaaru:
I see that checks for two spaces maybe could be good add vim tags in our specs file to ensure the logic matching that rules. This could be done directly in the three later.
Second, if it works we could add the same logic to the other helm templates ;)

geaaru:
Thanks

